Issue #12, opened 2021-04-03
Create a pre-curved rod


emersonkt:
Hello PyElastica,

I am a student and I am trying to apply the library 'PyElastica' in my project for a rod with a pre-curvature. But I could only create a straight rod. Could you please advise me if there is a way to make a pre-curved rod?

Thank you in advance for your help and congratulations for this beautiful work !

Émerson

armantekinalp:
Hi @emersonkt 

There are couple of things you need to do before creating a curved rod:
1) You need to create node positions as an array of `(3,n_elem+1)` which contains the `x,y,z` coordinates (row wise) of the nodes. Note that number of nodes is number of element +1 .
2) You need to compute tangents vector. Since you have positions vector already (in step 1), procedure is straight forward.
```python
tangents = position[…,1:] - position[…,:-1]
```
Tangents are unit vectors so we should normalize them by dividing its length. In PyElastica we have a build in function to compute lengths of vectors. Import the following;
```python
from elastica._linalg import _batch_norm

tangents /= _batch_norm(tangents)
```
3) The last array you need is for directors. Shape of directors is `(3,3,n_elem)`. We call the rows of director as `d1`, `d2`, and `d3`. You have already computed the `d3` in step 2 (`d3=tangents`). Now you only need to compute `d1` or `d2`. For example if you know `d1` then `d2` is cross product of `d3` and `d1`.

At this point you know position and directors of the rod. Next step is to create the rod and send `position` and `directors` as `kwargs`;

```python
shearable_rod = CosseratRod.straight_rod (n_elem, 
                                          start, 
                                          direction, 
                                          normal, 
                                          base_length, 
                                          base_radius, 
                                          density, 
                                          nu, 
                                          youngs_modulus, 
                                          poisson_ratio=0.5, 
                                          position=position, 
                                          directors=directors)
```

Since rod is not straight and curved initially we should compute the rest curvature and strain. However PyElastica has already computed curvature and strain for you when `shearable_rod` created.  You just need to update `rest kappa` and `rest sigma` as follows.
```python
shearable_rod.rest_kappa[:] = shearable_rod.kappa[:]
shearable_rod.rest_sigma[:] = shearable_rod.sigma[:]
```
Now you have a curved rod.

emersonkt:
Thank you very much ! It worked well and I got the desired pre-curved rod.

[...]

PR #14, opened 2021-04-07
bugfix:factory function allocate

armantekinalp:
This commit fixes a bug in `allocate` factory function. Previously if
user provides rod directors in kwargs, allocate was not updating the rod
directors. This commit fixes that bug and adds a new unit test for
testing if allocate function correctly assigns user input directors.


armantekinalp:
This commit solves the issue #12

PR #44, opened 2022-01-18
Energy plots for butterfly case

tp5uiuc:
Having energy plots is a useful diagnostic to test the accuracy of our Cosserat rod implementation along with the time-stepping algorithm.


armantekinalp (commented):
LGTM. 

Why do you save plots as pdf instead of png? Is there any reason

tp5uiuc:
The CI builds fail on an unrelated test case. Is this a known issue?

tp5uiuc:
> LGTM.
> 
> Why do you save plots as pdf instead of png? Is there any reason

No reason, the last plot (`butterfly.pdf`) was also saved as pdf, hence I just used the same.

armantekinalp:
Also can we change PR direction instead of `master` to `update-v0.2.1` ? We will do new realease soon. What do you think @skim0119

skim0119:
It is probably safer to merge into the update branch.

skim0119 (approved):
LGTM

PR #51, opened 2022-02-06
Update cosserat_rod parameter documentation

nmnaughton:
At some point, the Cosserat rod class had its documentation stripped. I think adding it back would be useful. I regularly check the v0.0.2 version of the docs for these details when I forget the syntax of a particular rod attribute.

### DIFF
--- a/elastica/rod/cosserat_rod.py
+++ b/elastica/rod/cosserat_rod.py
@@ -69,7 +69,6 @@ class CosseratRod(RodBase):
         damping_forces,
         damping_torques,
     ):
-        
         """
         Cosserat Rod class. This is the preferred class for rods because it is derived from some
         of the essential base classes.
--- a/elastica/rod/cosserat_rod.py
+++ b/elastica/rod/cosserat_rod.py
@@ -30,6 +30,91 @@ def _get_z_vector():
 
 
 class CosseratRod(RodBase):
+    """
+    Cosserat Rod class. This is the preferred class for rods because it is derived from some
+    of the essential base classes.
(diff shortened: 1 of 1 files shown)


nmnaughton:
Took a hot second to remember how exactly to get the doc string to show up. The proposed version is available here: https://docs.cosseratrods.org/en/update_cosserat_rod_doc/documentation.html while the current version without the info is here https://docs.cosseratrods.org/en/latest/documentation.html

In terms of ease of accessing useful information, I think it is better to just have it available.

elastica/rod/cosserat_rod.py
```diff
@@ -30,6 +30,91 @@ def _get_z_vector():
 
 
 class CosseratRod(RodBase):
+    """
+    Cosserat Rod class. This is the preferred class for rods because it is derived from some
+    of the essential base classes.
+    Although the attributes of this CosseratRod class are inherited from its parent classes, for convenience and easy access, the variable
+    names are given below.
```
armantekinalp, line 37: I think this statement is not accurate for the current code. Since `RodBase` does not have any attributes or methods (other than `__init__` )

armantekinalp (approved):
Looks good to me, I have one suggestion.

elastica/rod/cosserat_rod.py
skim0119: I agree with Arman. Maybe this explanation is not necessary. We don't have any other rod than CosseratRod at this point.

nmnaughton: I agree. I removed the relevant sentence.

skim0119 (approved):

Issue #36, opened 2022-01-06
Pyelastica API changes


skim0119:
<a href="https://github.com/armantekinalp"><img src="https://avatars.githubusercontent.com/u/53585636?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [armantekinalp](https://github.com/armantekinalp)**
_Monday Apr 13, 2020 at 16:07 GMT_
_Originally opened as https://github.com/GazzolaLab/elastica-python/issues/73_

----

These comments are from @tp5uiuc regarding the API of Elastica. I think they are good comments and we should try to accomplish these changes. @tp5uiuc and @nmnaughton please go over the below list and add your comments. Then after we finalize our discussion, I will start making changes.

- [x] #168
- [x] Changing `OneEndFixedRod` to `OneEndFixedBC` . (#55)
  - [x] #174
- [x] <s>Changing `callback functions` to `observer functions`</s>
- [x] <s>Changing `CallBackBaseClass` to `AbstractObserver` or `ObserverBaser`.</s>
  - @skim449: I'm scratching this out, because I think `Base` class is a somewhat common trend in many Python packages I'm seeing. There is really no community standard for this other than OOP languages. The word `abstract` could be too restrictive.
- [x] Reducing import statements, we can do
-- `from elastica.timesteppers import integrate, PositionVerlet`
-- `from elastica.wrappers import Connections, Constraints`
-- `from elastica import *`

## Other Refactor Request

-- Dissipation constant name to include `mass`
- [x] ExponentialDamping -> `AnalyticalLinearDamping` #163

skim0119:
<a href="https://github.com/skim0119"><img src="https://avatars.githubusercontent.com/u/3798023?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [skim0119](https://github.com/skim0119)**
_Monday Dec 27, 2021 at 02:16 GMT_

----

If we are going to refactor these, we should do it as soon as possible. To have a smooth transition, we can keep soft-reference link and have deprecated warning. In the version after, we cut-off the reference and throw deprecated error.

- I think callback name is okay, but maybe it is not common to have one callback for each system(rod).
- I think the word 'Base' is also fine. I'm seeing many ML python libraries use this convention. Only problem is that it doesn't force users to override the method. If this is needed, we should use `abstract` from `abc`.
[...]

PR #54, opened 2022-02-08
Update and generalize fixed rod BC

nmnaughton:
This partially addresses one of the comments in #36 regarding the `OneEndFixedRod` BC and also works to generalize the BC overall. There are three parts to the change. 

1) In the constraints wrapper, currently the 'constrained_position_idx' argument gets popped from the kwargs dict. If we change this to `get` then this idx argument continues onto the boundary condition and allows the boundary condition to associate the position and directors with specific idx in the rod.  @tp5uiuc I think you coded this originally, is there a reason we should be aware of why using `get` instead of `pop` would be advised?

2) Using the additional arguments from (1), any combination of nodes and elements can now be fixed in place by associating the idx and positions/directors info. Currently, I have added changes to `OneEndFixedRod` BC and also created an identical `FixedRodBC`. This is likely not the best solution. I am interested in hearing how people think we should proceed. Do we want to depreciate `OneEndFixedRod`? Is there a better naming scheme we could use?

3) I also included a separate BC (`FixedNodeBC`) where only the node is fixed. This would be a 'pinned' BC where the rod can still rotate around the point. 

The most useful part of this change (to me at least) is that one can now fix either end (or both) of a rod. 

I think one question for discussion is what should the name of the BCs be. Currently, there are `OneEndFixedRod`, `FixedRodBC`, and `FixedNodeBC`.  I like `FixedRodBC` and `FixedNodeBC`, but is there a better name for `FixedRodBC` ?  It should idealy denote that both an element and node is bing fixed. Maybe `FixedSectionBC`? Not sure...

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,17 +15,17 @@ import os
 import sys
 import sphinx_rtd_theme
 
-sys.path.insert(0, os.path.abspath("../"))
+sys.path.insert(0, os.path.abspath('../'))
 
 
 # -- Project information -----------------------------------------------------
 
-project = "Elastica"
-copyright = "2022, Gazzola Lab"
-author = "Gazzola Lab"
+project = 'Elastica'
+copyright = '2022, Gazzola Lab'
+author = 'Gazzola Lab'
 
 # The full version, including alpha/beta/rc tags
-release = "0.2.1"
+release = '0.2.1'
 
 
 # -- General configuration ---------------------------------------------------
@@ -34,40 +34,35 @@ release = "0.2.1"
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    "sphinx.ext.autodoc",
-    "sphinx.ext.napoleon",
-    "sphinx.ext.viewcode",
-    "sphinx_rtd_theme",
-    "sphinx.ext.mathjax",
+    'sphinx.ext.autodoc',
+    'sphinx.ext.napoleon',
+    'sphinx.ext.viewcode',
+    'sphinx_rtd_theme',
+    'sphinx.ext.mathjax',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
-templates_path = ["_templates"]
+templates_path = ['_templates']
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = []
 
-autodoc_default_flags = [
-    "members",
-    "private-members",
-    "special-members",
-    "show-inheritance",
-]
+autodoc_default_flags = ['members',  'private-members', 'special-members',  'show-inheritance']
 
 source_parsers = {
-    ".md": "recommonmark.parser.CommonMarkParser",
+   '.md': 'recommonmark.parser.CommonMarkParser',
 }
-source_suffix = [".rst", ".md"]
+source_suffix = ['.rst', '.md']
 
-master_doc = "index"
+master_doc = 'index'
 # -- Options for HTML output -------------------------------------------------
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "sphinx_rtd_theme"
+html_theme = 'sphinx_rtd_theme'
 
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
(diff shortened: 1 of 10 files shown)


skim0119:
@nmnaughton Could you redirect the PR to `update-0.3.0`, and remove the branch `update_fixed_BC`? I think any change in this PR should be included in the next version release. If you want a faster update, we can create `update-0.2.2` instead.

https://github.com/GazzolaLab/PyElastica/blob/master/CONTRIBUTING.md#project-workflow

skim0119:
@nmnaughton Hi, 👋  I'm going over this PR, and it appears to me that there is more than one place to make the modification, especially because the change involves the backend. I understand your opinions above and I agree with your points, but we should be careful in modifying the `constraint` part.

A couple of identified issues:
- The design choice of the `wrapper` and `finalize` operation is to distinguish the `model definition` from `execution`. The indices are not supposed to be visible in BC.
- `get`->`pop` is not the right approach. In fact, we specifically use `pop` to ensure the number of given indices for position and director match (line 197 in [constraint wrapper](https://github.com/GazzolaLab/PyElastica/blob/master/elastica/wrappers/constraints.py#L197)).
  - Hence, `_kwargs` is not passed to BC. In fact, both `_args` and `_kwargs` must be empty to run successfully, Otherwise, the above error would appear. (I understand it is a strange way.)

Here is my suggestion. ☕ 
- Save the branch and code in your fork for now. The new BC implementations look nice, and we should use them.
- Make the git-issue with the above problems that you would like to address.
- We can discuss the details on what is the best way to implement it.

nmnaughton:
Hi @skim0119, for the issues you raised

1) The point of this PR is to address precisely this fact.  While perhaps a design choice, I believe it was a mistake as it requires the BC to be hardcoded regarding the index. Any displacement BC inherently requires knowledge of which index the BC is being applied to, so we need to pass that info on somewhere. While the 2 displacement BC we have implemented (Helical and fixed end) do not need this idx info, any generalized class of displacement BC will. I regularly run into this issue and have to write my own BC to bypass this. 

2) We currently drop fixed_element_idx and fixed_directors_collection from the arguments list, why don't we just make them required in __init__? If you don't need them, you don't have to use them (for example, the Helical BC doesn't need to), but then they are available as needed. I have a commit I can push later that implements this approach and it seems like a clean solution. While that would be a breaking change for any custom BC people have, we could easily handle it using a depreciation notice.

skim0119:
> 1. The point of this PR is to address precisely this fact.  While perhaps a design choice, I believe it was a mistake as it requires the BC to be hardcoded regarding the index. Any displacement BC inherently requires knowledge of which index the BC is being applied to, so we need to pass that info on somewhere. While the 2 displacement BC we have implemented (Helical and fixed end) do not need this idx info, any generalized class of displacement BC will. I regularly run into this issue and have to write my own BC to bypass this.

Yes, I agree. This part of the user interface needs an update.

> 2. We currently drop fixed_element_idx and fixed_directors_collection from the arguments list, why don't we just make them required in **init**? If you don't need them, you don't have to use them (for example, the Helical BC doesn't need to), but then they are available as needed. I have a commit I can push later that implements this approach and it seems like a clean solution. While that would be a breaking change for any custom BC people have, we could easily handle it using a depreciation notice.

The issue is that **init** is not really what the user is exposed to, instead users are supposed to use 'using'. I think the biggest mistake here is that the name `constrained_position_idx` sounds like we give indices where the constraints are happening, but the actual constraint definition doesn't depend on that. 😭 

## Sugestion :coffee:

How about we change the syntax such that if you pass some arbitrary index that is not named `constrained_position(director)_idx`, they are passed to **init**. You can use and save them for later. For example,

```py
class CustomBC(FreeRod):
    def __init__(self, middle_indices:list[int]):
        FreeRod.__init__(self)
        self.middle_indices = middle_indices
        self.fixed_positions = self.get_system.position_collection[:, middle_indices]

    def constrain_values(self, rod, time): # Maybe rod can be replaced to self.get_system later
        rod.position_collection[:, middle_indices] = self.fixed_positions
        ...

...

simulator.constrain(rod).using(
    CustomBC,
    middle_indices=(6,7)
)
```

In this way, we don't need any deprecation warnings, as the previous features are all preserved. Passing arbitrary index is probably working even in the current version, but we just need `self.get_system` functionality.

If you already have a suggestion implemented as code, please link them 😄.

skim0119:
After talking to @nmnaughton and @tp5uiuc, I decided to change the current structure of constraint, so that indices are passed to the constraint. It looks like this feature was intentionally removed in the past, but it probably makes more sense to bring it back. Ill make the modification and include in the next version.

skim0119 (approved):
I'm merging these commits into wip_constraint. I'm going to continue on this problem.

PR #56, opened 2022-02-10
Update on constraint module; Add generalized constraint class

### LINKED ISSUE #55: [Dev Note] Boundary condition issues.
## Boundary condition model

Follow from #54. We are going to patch some patches to the current boundary condition implementation.

- [x] Enable boundary condition to be available at any specified node.
- [x] Expose system in each wrapper
- [x] Refactor names #36

skim0119:
## Dev Note

Related to #54 , the current `constraint` component does not pass the indices to the class methods. As a result, it is difficult to implement constraint or boundary conditions at the middle of the rod. I talked to @tp5uiuc, and this is definitely not intended behavior.

To fix the issue and improve future functionality, I'm exposing `system(rod)`, `position indices`, and `director indices` to the constraining/BC class. This should make BC implementation more flexible.

In addition, I refactored @nmnaughton 's implementation on more general BC: `FixedRodBC` and `FixedNodeBC` (#54). I added an example script on how to use new constraint `FixedConstraint`. **The PR also includes test codes**.

## How to use `FixedConstraint`

image

## Update

- `ConstraintBase` abstract class is added to provide template for all constraint classes.
  - People who inherited `FreeRod` should not get errors, but a deprecation warning should appear.
- Properties for `system`, `position_indices`, and `director_indices` are added.
- `FreeRod` and `OneEndFixedRod` are refactored to `FreeBC` and `OneEndFixedBC`. I made an alias and deprecation note so that people have time to notice the change and convert their code.
- Boundary condition `FixedConstraint` with documentation.
- Test code for `FixedConstraint` is added and tested.
- Test code for the `ConstraintBase` class is added and tested.
- Documentation is added for new classes.

resolve #55


elastica/boundary_conditions.py
```diff
@@ -1,70 +1,135 @@
...
+        Attributes
+        -------
+        system : RodBase
+        node_indices : None or numpy.ndarray
+        element_indices : None or numpy.ndarray
 
-    def constrain_values(self, rod, time):
+    """
+
+    _system: Type[RodBase]
```
armantekinalp, line 42: I think this will cause problems if we try to use BC for rigid bodies. Have you tested them @skim0119 , since rigid bodies are derived from `RigidBodyBase`

elastica/boundary_conditions.py
```diff
@@ -1,70 +1,135 @@
...
-
-        Returns
-        -------
-
         """
         pass
 
-    def constrain_rates(self, rod, time):
+    @abstractmethod
+    def constrain_rates(self, rod: Type[RodBase], time: float) -> None:
```
armantekinalp, line 94: why did we change from `system` to `rod`? We can use BC for rigid bodies as well.

elastica/boundary_conditions.py
```diff
@@ -152,7 +223,178 @@ def compute_constrain_rates(velocity_collection, omega_collection):
...
+            elif isinstance(data, np.ndarray) and data.shape == (
+                3,
+                3,
+            ):
+                dir.append(data)
+            else:
+                # TODO: This part is prone to error.
+                break
+        self.fixed_positions = np.array(pos)
+        self.fixed_directors = np.array(dir)
```
armantekinalp, line 282: An idea maybe is  hiding all these from user and handle everything inside the `_Constraint` class.

armantekinalp (changes requested):
I think everything looks good. I have some questions, could you please address them @skim0119

elastica/boundary_conditions.py
skim0119: We should refactor. All constrain_rates and constrain_values in current version `v0.2.1` said `rod`. I think eventually we should remove passing `rod/rigid-body` from outside, instead just call `get_system`.

elastica/boundary_conditions.py
skim0119: Good point. I changed to `Union[Type[RodBase], Type[RigidBodyBase]]`. In the future, we should have a shared base class for all `elastica-objects`. [e53c376](https://github.com/GazzolaLab/PyElastica/pull/56/commits/e53c376c7e690456de5818de1f8c51d8cee5dfa6)

elastica/boundary_conditions.py
```diff
@@ -1,70 +1,144 @@
...
+    Constraint class must inherit BaseConstraint class.
 
-    def __init__(self):
-        """
-        Free rod has no input parameters.
-        """
-        pass
+        Attributes
+        -------
+        system : RodBase
```
armantekinalp, line 37: Rigidbodybase

elastica/boundary_conditions.py
skim0119: Other functions' `__init__` include saving initial position/directors (`HelicalBucklingBC`, `OneEndFixedBC`). If we want to include this step in all constraints, we can add this inside the `_Constraint` class, but that would need a change in all implementations. Maybe we can refactor that in ` v0.3.0`. I wanted to keep this update `v0.2.1 -> v0.2.2`  to be soft.

armantekinalp: Okay then lets target v.0.3.0 for that.

armantekinalp (approved):
LGTM, one small documentation comment.

elastica/boundary_conditions.py
skim0119: fixed

PR #58, opened 2022-02-13
Porting getting started guide from website to docs page

nmnaughton:
I ported over the getting started guide from the website to the docs page per #49. It uses the new `myst_parser` version of the docs which doesn't look like it has been merged with `master` so I think this is the best branch to PR too.


skim0119 (approved):
Looks good to me

PR #64, opened 2022-03-05
[Core] Wrapper Redesign

skim0119:
# Wrapper structure redesign

I'm trying to address the below comments. This patch tries to fix the code structure accordingly. I hope this change makes the `wrapper` easier to understand for newcomers, and makes the documentation process nicer. The functionality should not have changed (All the pytest passes), because I just reorganized the structure.

https://github.com/GazzolaLab/PyElastica/blob/dd1bff433e7789914939e11eb6db90cfd60eee71/elastica/wrappers/base_system.py#L134-L144

## Discussion

> Note: I'm gonna use the term `base` to refer `BaseSystemCollection`, and `components` to refer to others (`Forcing`, `Constraints`, etc).

We have a conflict of design in our code structure. Fundamentally, we want to adapt the _Mixin_ paradigm as such
```py
class Simulator(BaseSystemCollection, Constraints, Forcing):
    pass

simulator = Simulator()
```
This design allows us to add/remove different modules easily, as long as we maintain the interface between components well. I think the problem comes because of the way each component is implemented: they resemble an _inheritance_ structure too much. For example, they share a similar structure overall, such as having `_finalize`, `__call__`, private wrapper class, `using`, etc. However, they will override themselves if we combine them to make _Mixin_ class, and this overriding is not part of our design choice.

The conflict of using `Mixin` and `OOP-inheritance` together shows in different places.

### 1. `self` in calling instance method??

https://github.com/GazzolaLab/PyElastica/blob/dd1bff433e7789914939e11eb6db90cfd60eee71/elastica/wrappers/base_system.py#L161-L182

Here, the first parameter of `feature`, which is some function in each component, needs `self` in their first parameter. This means they are used as a _static method_, even they are _instance method_!!  This is because functions in components are overriding themselves, and we need to backtrace which `feature` we want to call. Any developer will need to stop here and figure out what is going on.

### 2. Reflection

https://github.com/GazzolaLab/PyElastica/blob/dd1bff433e7789914939e11eb6db90cfd60eee71/elastica/wrappers/base_system.py#L138-L143

Here, the function `get_methods_from_feature_classes` collect methods in components using string comparison. Similar to above, we are forced with this functionality not because we want to collect functions, which should be a simple job of what `super()` should be doing, but because we lost overridden methods. As a result, we lose the debuggability of the code after this point, and we need over usage of the reflection technique to correct it.

### 3. Missing Abstract/Base for component

We have components (`Forcing`, `Constraints`, `Connection`, `Callback`) that share similar properties as mentioned above, but we cannot have a single abstract function that enforces the structure. This is because we cannot call `super()` and backtrace the MRO of the class.

## New Structure Suggestion

I'm suggesting a simple solution: let's drop the inheritance (it is not inheriting anything anyway). The method `finalize` and `call` look similar in each component, but they essentially serve different purposes. We cannot rely on `super()` because the order of execution is not well-synchronized. Keeping the method name the same across the component is confusing, error-prone, and dangerous without a proper inheritant structure. Most importantly, I don't see it as necessary.

_Mixin_ is powerful when a functionality needs to be added in to existing class, just like how Django extend the functionality using _Mixin_. Lets just do that: each component provide new functionality.

### What are we doing in the `finalize()`?

Essentially, the way I see the key functionality of `finalize` step is to allocate different component-methods into `synchronize`, `constrain_values`, `constrain_rates`, `finalize`, and `apply_callbacks` group. The group of methods is executed at different parts of the time stepper. Below is the diagram.

image

Previously, we rely on `get_methods_from_feature_classes` to draw the connection arrow (blue arrow). I changed it so that the connection arrow (blue arrow) is registered by each component. This reduces the need of using string-comparison to organize the method. Furthermore, it should be the extendable design of choice, in case we want to add another feature group.

### Rename \_\_call\_\_

This dunderscore function makes the class `Callable`, but I don't see why: we never call simulator itself. Furthermore, this is dangerous because `__call__` is overridden for each components, and we don't know which one is called. I'm renaming the method.

## Changes

- Add an error message when a user tries to finalize twice. A `finalize_flag` is added to detect additional finalize calls. Previously, calling finalize twice gave an error, but the error message was not so useful to detect.
- Modify the finalize and wrapper structure.
- Modify test cases

## Note

- The PR is ready for `update-0.2.2`, but I wouldn't mind pushing to `update-0.3.0` to test more.

### DIFF
--- a/elastica/wrappers/base_system.py
+++ b/elastica/wrappers/base_system.py
@@ -51,6 +51,9 @@ class BaseSystemCollection(MutableSequence):
         # from Controllers, Environments which are
         # tacked on to the SystemCollection in a sim.
         self._features = NotImplemented
+        # Flag Finalize: Finalizing twice will cause an error,
+        # but the error message is very misleading
+        self._finalize_flag = False
 
     def _check_type(self, sys_to_be_added):
         if not issubclass(sys_to_be_added.__class__, self.allowed_sys_types):
@@ -131,6 +134,8 @@ class BaseSystemCollection(MutableSequence):
 
         """
 
+        assert self._finalize_flag is not True, "The finalize cannot be called twice."
+
         # FIXME: This is probably the most bizarre way to collect the functions. It is definitely
         # impressive and even surprising that it is working, but it is far from readable and maintainable.
         # The code is difficult to debug, because the behavior is un-interpretable except during run-time.
@@ -160,6 +165,8 @@ class BaseSystemCollection(MutableSequence):
 
         for finalize in finalize_methods:
             finalize(self)
+
+        self._finalize_flag = True
 
     def synchronize(self, time):
         # Calls all , connections, controls etc.
--- a/elastica/wrappers/base_system.py
+++ b/elastica/wrappers/base_system.py
@@ -41,12 +41,12 @@ class BaseSystemCollection(MutableSequence):
     """
 
     # Collection of functions. Each group is executed as a collection at the different steps.
-    # Each component (Forcing, Connection, etc.) registers the executable (callable) function 
+    # Each component (Forcing, Connection, etc.) registers the executable (callable) function
     # in the group that that needs to be executed.
-    _feature_group_synchronize: Iterable[Callable[[float],None]] = []
-    _feature_group_constrain_values: Iterable[Callable[[float],None]] = []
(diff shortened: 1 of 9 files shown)


armantekinalp (approved):
Very good points. Now interface looks cleaner and easier to understand and extend. LGTM we can merge.

PR #70, opened 2022-03-12
Add functions to compute topological quantities

skim0119:
# Topological Quantities: Link, Writhe, and Twist 

Knot-theory module added to Cosserat-rod as _mixin_. It extends the Cosserat-rod class to compute link, writhe, and twist (LWT).

## Tasks before the merge

- [x] Unittest
- [x] Enable/allow computing functions to be used in post-processing.
- [x] In-line documentations
- [x] Documentation page for LWT description
  - [x] Also add explanation on how to batch-process.
  - [x] Link examples: `Solenoids`, `Plectomenes`
  - [x] <s>Advanced documentation on `type_of_additional_segment`</s>
    - Separate collection of reference will include papers related to LWT. Additional table and description is added in KnotTheory docstring.

## Design purpose ☕ 

- The implementation is in a separate module because LWT property is not tied to the Cosserat rod only.
- The protocol is given to check the compatibility of mixin. Here, we check `RodBase` and `KnotTheoryCompatibleProtocol`. The protocol `KnotTheoryCompatibleProtocol` checks if necessary properties (i.e. collections) exist.

## Note 📔 

- These computations are relatively expensive and much more efficient to do in the post-processing stage. It is included as rod-property, because users might want to use LWT in their application, but not recommended to use in every iteration (e.g. callback). We should state this clearly and provide examples of expected usage in the documentation.
- I'm targeting this PR on update `v0.3.0` because there is no need to rush, but we can try to include it earlier if completed.


armantekinalp:
I run the example cases for LWT, here are the results.

For plectoneme:
link_twist_writhe_plectoneme

For solenoid:
link_twist_writhe_solenoid

armantekinalp (approved):
LGTM

PR #71, opened 2022-04-10
Patch: Rod constructor warning messages

skim0119:
- The logging level of some messages is lowered (warning->info) in order to avoid unnecessary logging.
  - Technically the warning should be given when both `poisson_ratio` and `shear_modulus` are given. In this case, a warning should say `poisson_ratio` will be ignored.
  - In any other case, we just need to `log` what happened. I don't think it is necessary to throw a warning.
- Refactored the name: too many `kwargs` lowers readability.
- Refactored `allocate`


armantekinalp (approved):
LGTM lets merge

PR #72, opened 2022-04-17
Muscular snake implementation added as a new example

armantekinalp:
Features

- How to use parallel connection.
- How to use friction module.
- Custom muscle force module.

After getting final results, I will publish here and then we can merge.


examples/MuscularSnake/get_connection_vector.py
```diff
@@ -0,0 +1,54 @@
+import numpy as np
+# Join the two rods
+from elastica._linalg import (
+    _batch_norm,
+    _batch_matvec,
+)
+
+
+
+def get_connection_vector_straight_straight_rod(
```
skim0119, line 10: Don't we have this in some experimental module?

[here](https://github.com/GazzolaLab/PyElastica/blob/update-0.2.3/elastica/experimental/connection_contact_joint/parallel_connection.py)

examples/MuscularSnake/muscular_snake.py
```diff
@@ -0,0 +1,400 @@
+__doc__ = """Muscular snake example from Zhang et. al. Nature Comm 2019 paper."""
+import sys
+
+sys.path.append("../../")
```
skim0119, line 4: remove?

skim0119 (changes requested):

examples/MuscularSnake/muscular_snake.py
armantekinalp: ### 
For now lets keep it after creating separate example repo, we can remove this. I think otherwise there will be path problem

examples/MuscularSnake/get_connection_vector.py
armantekinalp: This is function computes the connection vector for specified index of rod one. I will update this [one](https://github.com/GazzolaLab/PyElastica/blob/update-0.2.3/elastica/experimental/connection_contact_joint/parallel_connection.py)

skim0119 (approved):
LGTM

PR #84, opened 2022-05-16
Blacken example files (#83)

bhosale2:
@armantekinalp as per discussion, this PR provides blacken option for examples.


armantekinalp (approved):
LGTM

PR #88, opened 2022-05-17
(#78) fix compute shear energy function typo

### LINKED ISSUE #78: Typo in `compute_shear_energy` method of Cosserat Rod.
Although the code to compute shear energy function is correct, naming of one of the variables is wrong. `shear_internal_torques` is not correct and misleading, we should call this variable as `shear_internal_forces` . 

https://github.com/GazzolaLab/PyElastica/blob/fee87b9da22e310ff925c16fdc839bf8405c51a4/elastica/rod/cosserat_rod.py#L516

bhosale2:
Fixes #78. (quick one)


skim0119 (approved):
LGTM

PR #114, opened 2022-06-16
Refactor internal dissipation as external addon damping module 

### LINKED ISSUE #112: Factor out internal dissipation as external addon dissipation module
**Is your feature request related to a problem? Please describe.**
Currently, the internal dissipation used to stabilize PyElastica simulations is a part of the Cosserat rod governing equations. This leads to the following 2 issues:
1. In over/critically damped cases, this can lead to a severe limitation on the timestep size, slowing down simulations.
2. Does not allow for a cleaner way to simulate an undamped rod, other than setting `nu = 0` and crunching the damping steps for no reason.

**Describe the solution you'd like**
Following discussion with @armantekinalp, the dissipation step can be factored out of the Cosserat rod equations, and treated analytically, resulting in an exponential damper dissipation module, which is 
1. Cleaner add-on module to undamped Cosserat rod.
2. Imposes no timestep restrictions on the simulator.
3. Allows for cleaner testing for other dissipation schemes (like filtering)

bhosale2:
Fixes #112 by providing an alternative add-on damping module (mixin class) for internal dissipation of Cosserat rods, with the following advantages:

1. Damping can be added as a Mixin class when it is needed for a rod, thus making damping modular.
2. The analytical version of the original internal damper `ExponentialDamper` is unconditionally stable, and removes any timestep restrictions coming in from damping.
3. With the incoming modularity, other damping models can be conveniently implemented.

Additionally, a `DeprecationWarning` is added when the damping constant is provided in the rod constructor (soon the option will be removed in the following release-0.4).

The examples will be refactored (with help from @armantekinalp), along with confirmation, that the dynamics of the individual cases stay the same:

- [x] axial stretching
- [x] binder stuff?
- [x] butterfly
- [x] continuum flagella
- [x] continuum snake
- [x] Experimental cases - parallel connection
- [x] flexible swinging pendulum
- [x] friction validation cases
- [x] helical buckling
- [x] joint cases
- [x] muscular flagella
- [x] muscular snake
- [x] restart example
- [x] rod-rod contact cases
- [x] rod self contact cases
- [x] timoshenko

@armantekinalp after finishing the confirmations, we should also check if we can raise timesteps in these cases.


elastica/dissipation.py
```diff
@@ -0,0 +1,117 @@
...
+]
+
+from typing import Type, Union
+import numpy as np
+from abc import ABC, abstractmethod
+from elastica.rod import RodBase
+from elastica.rigidbody import RigidBodyBase
+
+
+class DamperBase(ABC):
```
skim0119, line 14: @bhosale2 You sure ABC is the right one you wanna use here?

elastica/dissipation.py
```diff
@@ -0,0 +1,117 @@
...
+    All damper classes must inherit DamperBase class.
+
+
+        Attributes
+        ----------
+        system : RodBase or RigidBodyBase
+
+    """
+
+    _system: Union[Type[RodBase], Type[RigidBodyBase]]
```
skim0119, line 28: I think we should have separate typing alias in separate file. Could you maybe make `elastica/typing.py` and have `SystemType: Union[Type[RodBase], Type[RigidBodyBase]]`?

elastica/rod/factory_function.py
```diff
@@ -192,6 +192,18 @@ def allocate(
     mass[1:] += 0.5 * density * volume
 
     # Set dissipation constant or nu array
+    warnings.warn(
```
skim0119, line 195: Please use `logging` module instead of `warnings`. It will become more useful when we want to integrate with configuration modules, like `sacred`.

examples/AxialStretchingCase/axial_stretching.py
```diff
@@ -87,6 +88,16 @@ class StretchingBeamSimulator(BaseSystemCollection, Constraints, Forcing, CallBa
     EndpointForces, 0.0 * end_force, end_force, ramp_up_time=1e-2
 )
 
+# add damping
+damping_constant = 1.0
+dl = base_length / n_elem
+dt = 0.01 * dl
+stretch_sim.dampen(stretchable_rod).using(
+    ExponentialDamper,
```
skim0119, line 96: Open for discussion. Do we wanna use `Damper`? Wouldn't it have the potential to mislead what it is? For example, we can have external `damper` implemented as force, etc. This is more `internal` feature, I feel like.

skim0119 (changes requested):
I like the change, but we should be careful on how we define the terms: `dissipation` vs `damping`.  Can you include a page in the documentation?

elastica/dissipation.py
bhosale2: The idea here was to design the `damping` module the same way as `constraints`, where an abstract class with an abstract method `dampen_rates()` needs to be defined. @skim0119 if you feel I'm missing something here following the way things are done in `pyelastica` let me know. Or better please suggest the alternative you have in mind.

examples/AxialStretchingCase/axial_stretching.py
bhosale2: I don't believe that `damping` is an internal feature since the original Cosserat rod equations don't have this term. Rather just like forces/torques, it's an external to-be documented feature (available and visible to the users) that's added to the simulator to numerically `soften` or stabilize the simulations. Additionally, the presence of documentation, which will be added soon by @armantekinalp  and I will further avoid any potential confusion that may arise. Finally, the damper you are referring to can be appropriately named as `DampingForce` and added through the `Forcing` class, thus avoiding any potential clash/confusion with this module.

armantekinalp: I agree with @bhosale2 on this point. Cosserat Rod does not have damping internally build in so having it as a separate  module makes more sense. Also with documentation it will be more clear the details and how to use `damping` module.

bhosale2:
@armantekinalp here is the checklist for updating docs for this module:
 
- [ ] Update `index.rst` to reflect damping.
- [ ] Add `damping.rst` to `docs/api`.
- [ ] Update `workflow.md` in `docs/guide`.

armantekinalp:
I add the documentation for the new `dissipation` module. @bhosale2 and @skim0119 please go over. After the review I think PR is ready to merge.

docs/index.rst
```diff
@@ -53,6 +53,7 @@ If you are interested to contribute, please read `contribution-guide`_ first.
    
    api/rods
    api/rigidbody
+   api/damping
```
skim0119, line 56: `damping` below `time_steppers` maybe?

docs/requirements.txt
```diff
@@ -8,4 +8,5 @@ sphinx-autodoc-typehints
 myst-parser
 numpydoc
 docutils
+sphinx_rtd_theme
```
skim0119, line 11: why? we use book theme, right?

elastica/wrappers/damping.py
```diff
@@ -0,0 +1,156 @@
...
+        # Thus using lambda we iterate over the list of tuples and use rod number (x[0])
+        # to sort dampers.
+        self._dampers.sort(key=lambda x: x[0])
+
+    def _dampen_rates(self, time, *args, **kwargs):
+        for sys_id, damper in self._dampers:
+            damper.dampen_rates(self._systems[sys_id], time, *args, **kwargs)
+
+
+class _Damper:
```
skim0119, line 75: Rename to `_Damping`? Its suppose to be private class wrapper for `Damping`.

skim0119 (changes requested):
I'm still looking into the PR. Here is some minor requests to change.

skim0119:
Are we planning to include filtering as part of `dissipation.py`?

bhosale2:
@skim0119 yes that has been the plan from the start.

docs/index.rst
bhosale2: The reason for that is the Damping wrapper actually holds a list of the "dampers" initialized through the "_damper" class. Hence the naming. I'm not too stuck up on this and arman can change this if he feels like.

skim0119:
> @skim0119 yes that has been the plan from the start. 

Then I think we should give more proper name than 'dissipation', like 'magic.py' or 'accelerator.py' ;)

elastica/wrappers/damping.py
bhosale2: The reason for that is the Damping wrapper actually holds a list of the "dampers" initialized through the "_damper" class. Hence the naming. I'm not too stuck up on this and arman can change this if he feels like.

skim0119: Yeah, lets keep it consistent with other naming. Im looking into the code

bhosale2:
I think we want to publish it and make this feature available to everyone.

elastica/wrappers/damping.py
bhosale2: Okay, if this and the other changes are minor can you please take care of them?

docs/requirements.txt
armantekinalp: Documentation build was giving errors, that is why I included this? Are you sure this is not a related package we need?

armantekinalp:
I agree with @bhosale2 . This will benefit the community a lot and people should be able use this code sooner than later.

docs/requirements.txt
skim0119: I'll double check. It was removed with a reason.

bhosale2: I'm resolving this for now, since this is related to docs in general. We can do a quick fix later if needed.

elastica/wrappers/damping.py
armantekinalp: I think `_damper` make sense and consistent with other wrappers the for example `Forcing`. The main wrapper that calls all external force classes is called `Forcing` and individual external force objects are called `_ExtForceTorque` . Similarly we name wrapper that calls all `damper` classes as `Damping` and individual damper objects as `_Dampers`.

bhosale2:
@skim0119 please review it again with a fresh mind (one-shot preferable) and let's merge this soon since this will greatly benefit the users.

skim0119 (approved):
LGTM

PR #117, opened 2022-06-24
116 rigid body rod contact

### LINKED ISSUE #116: Rigid body - rod contact torque
**Describe the bug**
Contact module between rigid body and rod does not incorporate torques generated due to contact. 

**Expected behavior**
For all cases that rod and rigid body are not parallel, rigid body should start rotating after contact.

armantekinalp:
Fixes #116 

Previous implementation of rigid cylinder rod contact was not taking into account torques generated on cylinder due to contact. This PR changes the previous implementation to incorporate torques generated by contact forces that are applied away from the cg of cylinder. 

In order to validate implementation a new example simulation called `rod_cylinder_parallel_contact.py` is added where parallel rod and cylinder contact and exchange momentum. Simulation results are given below, where `rod_one` is a Cosserat Rod , `rod_two` is rigid cylinder.

rod_rigid_velocity

Blue rod is Cosserat Rod, Orange rod is Rigid Cylinder.

https://user-images.githubusercontent.com/53585636/175464853-cb154451-bb60-40f5-8cf6-c923799e66c3.mp4


bhosale2 (approved):
LGTM. But to show torques are acting, shouldn't we show an example where there is an inclined collision, and then rotation happens? What if the inclination angle is provided as a parameter to be set in the example case?

armantekinalp:
I will add one more example, where cylinder rotates after collision.

bhosale2:
@armantekinalp after you add the variable-angle collision case, no need to re-request, you can merge directly.

armantekinalp:
Inclined rod - cylinder contact 
rod_rigid_velocity

https://user-images.githubusercontent.com/53585636/176969262-3593632d-4eff-47e0-98ab-fc916497fdd4.mp4

PR #125, opened 2022-07-04
Fix #128:  save_every condition in ExportCallBack

### LINKED ISSUE #128: Bug for asynchronous specs of skip_step and save_every in ExportCallBack
**Describe the bug**
If `save_every` is not a multiple of `step_skip`, the condition `(current_step + 1) % self.save_every == 0` (https://github.com/GazzolaLab/PyElastica/blob/master/elastica/callback_functions.py#L205) will never be trigged and the data will never be saved, if the condition about the buffer size does not become active neither.

**To Reproduce**
This code snippet below has a time-step of `0.1`ms. It should save the data to the buffer every `20`ms and it should dump the buffer to the file every `1.0001`ms.
```py
dt = 1e-4
step_skip = 2e2
save_every = 10001
callback = ExportCallBack(step_skip, ".", "tempfile", save_every=10)
simulator.collect_diagnostics(stretchable_rod).using(
     callback, 
     step_skip=step_skip, 
     path=".", method="tempfile", 
     save_every=save_every,
)
```
However, in this case the `(current_step + 1) % self.save_every == 0`  will never be triggered, as it never reaches there as this time-step is skipped by `step_skip=2e2`. PR #125 would fix this.

**Environment**
- Python=3.10.5, pip=22.1.2 
- Numpy/numba version: 22.1.2
- PyElastica version: 0.2.3

**Expected behavior**
By specifying `save_every`, the data should be saved every `save_every` time-steps, irrespective of the chosen `skip_step` setting. 

**Screenshots**

**Additional context**

mstoelzle:
The previous implementation would require "skip_step" and "save_every" to be synchronous in order to trigger. Now, we instead count the number of time steps since the last save and trigger the save whenever this number is higher or equal than the "save_every" parameter

### DIFF
--- a/elastica/callback_functions.py
+++ b/elastica/callback_functions.py
@@ -155,7 +155,6 @@ class ExportCallBack(CallBackBaseClass):
 
         self.buffer = defaultdict(list)
         self.buffer_size = 0
-        self.steps_since_last_save: int = 0
 
         # Module
         if method == ExportCallBack.AVAILABLE_METHOD[0]:
@@ -185,7 +184,6 @@ class ExportCallBack(CallBackBaseClass):
         current_step : int
             simulation step
         """
-        self.steps_since_last_save += 1
 
         if current_step % self.step_skip == 0:
             position = system.position_collection.copy()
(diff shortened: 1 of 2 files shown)


skim0119:
@mstoelzle Hi, Thank you for your consideration in PyElastica.

Can you clarify the bug a little more, maybe specify the expected behavior?

Maybe I should have named and documented the variables more clearly; `skip_step` and `save_every` has two different purposes. The parameter `skip_step` is the number of step intervals to save the simulation. For example, if `dt=1e-4` and `skip_step=1000`, the data will save with the interval of `0.1` second. These data will be saved in the buffer dictionary.

On the other hand, `save_every` specify the number of timesteps before this dictionary is dumped into the file. For example, if `save_every=1e8`, the save-file will be generated every `1000` seconds of simulation. The purpose is to reduce the number of files generated during the simulation.

If you have a different idea to implement, we would appreciate it if you make an issue to discuss further.

elastica/callback_functions.py
```diff
@@ -184,6 +184,7 @@ def make_callback(self, system, time, current_step: int):
         current_step : int
             simulation step
         """
+
```
skim0119, line 187: Could you remove the blank change?

PR #126, opened 2022-07-04
Fix bug in checking of path for ExportCallback

mstoelzle:
If any path not equal to a directory is specified, the current implementation throws an error that the export path does not exist. However, it should also be possible to use filenames (instead of just dirnames) such as:
```python
path="dirname/filename"
```
and then the first batch of data should be for example saved as `dirname/filename_0.dat.npz`.


skim0119:
@mstoelzle

Thank you for the bug report. I see the problem you pointed out, but I don't think it solves the entire problem. The module does not have a test case for saving `numpy` or `pickle` format, which is probably why the bug slipped. I'm moving this to #127, and I'll update the fix in `v0.2.4`.

PR #130, opened 2022-07-11
Update ExportCallback

### LINKED ISSUE #127: Fix bug in checking of path for ExportCallback
<a href="https://github.com/mstoelzle"></img></a> **Issue by [mstoelzle](https://github.com/mstoelzle)**
_Originally opened as https://github.com/GazzolaLab/PyElastica/pull/126_

----

If any path not equal to a directory is specified, the current implementation throws an error that the export path does not exist. However, it should also be possible to use filenames (instead of just dirnames) such as:
```python
path="dirname/filename"
```
and then the first batch of data should be for example saved as `dirname/filename_0.dat.npz`.

skim0119:
##  Note

resolve #127 

- Fix bugs in saving
- Parameter `filename` and `directory` is added for clarity.
- New directory will be created if it does not exist.
- Added `get_last_saved_path` method for path check.
- Added `close` and `clear` instead of `__del__` which could be more vulnerable to other exceptions.
- Added more test cases.

### DIFF
--- a/elastica/callback_functions.py
+++ b/elastica/callback_functions.py
@@ -144,7 +144,9 @@ class ExportCallBack(CallBackBaseClass):
         # Assertions
         MIN_STEP_SKIP = 100
         if step_skip <= MIN_STEP_SKIP:
-            logging.warning(f"We recommend ({step_skip=}) at least {MIN_STEP_SKIP}")
+            logging.warning(
+                f"We recommend (step_skip={step_skip}) at least {MIN_STEP_SKIP}"
+            )
         assert (
             method in ExportCallBack.AVAILABLE_METHOD
         ), f"The exporting method ({method}) is not supported. Please use one of {ExportCallBack.AVAILABLE_METHOD}."
--- a/elastica/callback_functions.py
+++ b/elastica/callback_functions.py
@@ -129,7 +129,7 @@ class ExportCallBack(CallBackBaseClass):
             interval.
         filename : str
             Name of the file without extension. The extension will be
-            determined depend on the method. 
+            determined depend on the method.
         directory : str
             Path to save the file. If directory doesn't exist, it will
             be created. Any existing files will be overwritten.
@@ -152,7 +152,9 @@ class ExportCallBack(CallBackBaseClass):
 
         # Create directory
         if os.path.exists(directory):
-            logging.warning(f"Save file already exists in {directory}. Files will be overwritten")
+            logging.warning(
+                f"Save file already exists in {directory}. Files will be overwritten"
+            )
         os.makedirs(directory, exist_ok=True)
 
         # Argument Parameters
@@ -215,7 +217,7 @@ class ExportCallBack(CallBackBaseClass):
             )
 
         if (
-            self.buffer_size > ExportCallBack.FILE_SIZE_CUTOFF
+            self.buffer_size > self.FILE_SIZE_CUTOFF
             or (current_step + 1) % self.file_save_interval == 0
         ):
             self._dump()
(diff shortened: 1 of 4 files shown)


armantekinalp (approved):
LGTM

elastica/callback_functions.py
```diff
@@ -111,52 +113,59 @@ class ExportCallBack(CallBackBaseClass):
...
-        path : str
-            Path to save the file. If directories are prepended,
-            they must exist. The filename depends on the method.
-            The path is not expected to include extension.
+        filename : str
+            Name of the file without extension. The extension will be
+            determined depend on the method.
+        directory : str
+            Path to save the file. If directory doesn't exist, it will
+            be created. Any existing files will be overwritten.
```
mstoelzle, line 134: This should probably say:

_Directory to save the file. If directory doesn't exist, it will be created. Any existing files in the directory will be overwritten._

elastica/callback_functions.py
```diff
@@ -111,52 +113,59 @@ class ExportCallBack(CallBackBaseClass):
...
             method in ExportCallBack.AVAILABLE_METHOD
         ), f"The exporting method ({method}) is not supported. Please use one of {ExportCallBack.AVAILABLE_METHOD}."
-        # TODO: Assertion is temporarily disabled. Should be fixed with #127
-        # assert os.path.exists(path), "The export path does not exist."
+
+        # Create directory
+        if os.path.exists(directory):
+            logging.warning(
+                f"Save file already exists in {directory}. Files will be overwritten"
+            )
```
mstoelzle, line 158: I think in the current version of the code, this is a bug. Because the if statement doesn't check if a **file** exists, but rather checks if a **directory** exists. Therefore, the warning is misleading. I would propose the following code:
```py
if not os.path.exists(directory):
           os.makedirs(directory) 
```

And then later when `file_path` is defined:
```py
if not os.path.exists(file_path):
             logging.warning(
                  f"Save file already exists in {directory}. Files will be overwritten"
              )
```

mstoelzle (commented):
I think there is a potential bug / misleading warning when checking the directory

elastica/callback_functions.py
skim0119: I don't see why the first change is needed. With current code, the directory will be created regardless of whether the directory exists. Adding an if statement makes the code redundant.

I don't think the second point is a bug. The file name will typically have numbers and extensions attached, and we don't want to throw a warning for every file we overwrite. The point here is that the directory already exists and gives a warning to the user that the previous data can be overwritten. It is up to the user whether to disregard the previous simulation or to change the directory name.

I agree that the current message can be misleading a little, and the documentation doesn't exactly describe what will happen. I'll change the warning message on the upcoming `v0.2.4` update.

elastica/callback_functions.py
skim0119: Okay, I'll clarify.

skim0119: [9e16e3d](https://github.com/GazzolaLab/PyElastica/commit/9e16e3d484d9c30673cc00897e5d06b091db08c4)

elastica/callback_functions.py
skim0119: [9e16e3d](https://github.com/GazzolaLab/PyElastica/commit/9e16e3d484d9c30673cc00897e5d06b091db08c4)

mstoelzle: I think commit https://github.com/GazzolaLab/PyElastica/commit/9e16e3d484d9c30673cc00897e5d06b091db08c4 mostly solves the mentioned issues. The only question I would pose is whether you would like to warn the user in the following case:
- The directory `my_folder` already exists, and it is either empty or there exists some file with the name `old_test_*`
- The user has specified as directory `my_folder` and as filepath `new_test`
I think in this case in might confuse the user to have a warning, as really nothing can be overwritten. It would be better to check whether any files of name `new_test_*` exists in the folder. If yes, then the warning definitely is necessary. Otherwise, it is not needed.

skim0119: Thank you for addressing your concern.

In my opinion, the `warning` is given just to let users to `note` that a previous simulation could exist. I don't think overwriting the file itself is technically an issue and **in most cases, it is desirable**. This is why I don't think having a warning at the overwriting is necessary. If the policy was to avoid overwriting, I would have prompted users if it is okay, or throw an error. Ultimately, keeping simulation data safe is the user's responsibility, and the user should be careful running the same simulation without saving previous runs.

As you said, it will throw a warning even if there is no saved data, but the fact that the folder exists could mean the script had been executed. Again, the warning is not so much about **overwriting**, but mentions potential repeated execution.

Besides practically, I would expect the user to copy the data directory to be `my_folder -> old_my_folder` instead of each file. I usually add `strftime` at the end of the folder name.

PR #135, opened 2022-07-14
Update FixedJoints: restoring spring-damped-torques, initial rotation offset

### LINKED ISSUE #131: Twisting torque is not constrained by FixedJoint and HingeJoint
## Describe the bug
The current implementation of the FixedJoint based a position spring does not constrain any twisting / torsional torques applied to a straight rod. This happens, because the current approach (essentially) computes an error between the tangential vector of rod 1 and tangential error of rod 2. As rotating around the z-axis / tangential vector does not change this vector, the twisting / torsion is not constrained by the FixedJoint.

In the example below, we have the classic example of two rods connected by a FixedJoint. They are both initially in straight configuration and we apply a UniformTorque to the second rod. For a twisting torque, we notice the faulty behaviour, while for a bending torque everything works as expected.

## To Reproduce
```py
__doc__ = """Fixed joint example, for detailed explanation refer to Zhang et. al. Nature Comm.  methods section."""

import matplotlib.pyplot as plt
import numpy as np
import sys

# FIXME without appending sys.path make it more generic
sys.path.append("../../")
from elastica import *
from examples.JointCases.external_force_class_for_joint_test import (
    EndpointForcesSinusoidal,
)
from examples.JointCases.joint_cases_callback import JointCasesCallback
from examples.JointCases.joint_cases_postprocessing import (
    plot_position,
    plot_video,
    plot_video_xy,
    plot_video_xz,
)


class FixedJointSimulator(
    BaseSystemCollection, Constraints, Connections, Forcing, CallBacks
):
    pass


fixed_joint_sim = FixedJointSimulator()

# setting up test params
n_elem = 10
direction = np.array([0.0, 0.0, 1.0])
normal = np.array([0.0, 1.0, 0.0])
roll_direction = np.cross(direction, normal)
base_length = 0.2
base_radius = 0.007
base_area = np.pi * base_radius ** 2
density = 1750
nu = 1e-1
E = 3e7
poisson_ratio = 0.5
shear_modulus = E / (poisson_ratio + 1.0)

start_rod_1 = np.zeros((3,))
start_rod_2 = start_rod_1 + direction * base_length
start_cylinder = start_rod_2 + direction * base_length

# Create rod 1
rod1 = CosseratRod.straight_rod(
    n_elem,
    start_rod_1,
    direction,
    normal,
    base_length,
    base_radius,
    density,
    nu,
    E,
    shear_modulus=shear_modulus,
)
fixed_joint_sim.append(rod1)
# Create rod 2
rod2 = CosseratRod.straight_rod(
    n_elem,
    start_rod_2,
    direction,
    normal,
    base_length,
    base_radius,
    density,
    nu,
    E,
    shear_modulus=shear_modulus,
)
fixed_joint_sim.append(rod2)

# Apply boundary conditions to rod1.
fixed_joint_sim.constrain(rod1).using(
    OneEndFixedBC, constrained_position_idx=(0,), constrained_director_idx=(0,)
)

# Connect rod 1 and rod 2
fixed_joint_sim.connect(
    first_rod=rod1, second_rod=rod2, first_connect_idx=-1, second_connect_idx=0
).using(FixedJoint, k=1e5, nu=0, kt=5e3)

# Add forces to rod2
fixed_joint_sim.add_forcing_to(rod2).using(
    UniformTorques,
    torque=5e-3,
    direction=np.array([0., 0., 1.])
)


pp_list_rod1 = defaultdict(list)
pp_list_rod2 = defaultdict(list)


fixed_joint_sim.collect_diagnostics(rod1).using(
    JointCasesCallback, step_skip=1000, callback_params=pp_list_rod1
)
fixed_joint_sim.collect_diagnostics(rod2).using(
    JointCasesCallback, step_skip=1000, callback_params=pp_list_rod2
)

fixed_joint_sim.finalize()
timestepper = PositionVerlet()
# timestepper = PEFRL()

final_time = 10
dl = base_length / n_elem
dt = 1e-5
total_steps = int(final_time / dt)
print("Total steps", total_steps)
integrate(timestepper, fixed_joint_sim, final_time, total_steps)


def plot_orientation_vs_time(title, time, directors):
    from scipy.spatial.transform import Rotation as R
    quat = []
    for t in range(len(time)):
        quat_t = R.from_matrix(directors[t].T).as_quat()
        quat.append(quat_t)
    quat = np.array(quat)

    plt.figure(num=title)
    plt.plot(time, quat[:, 0], label="x")
    plt.plot(time, quat[:, 1], label="y")
    plt.plot(time, quat[:, 2], label="z")
    plt.plot(time, quat[:, 3], label="w")
    plt.title(title)
    plt.legend()
    plt.xlabel("Time [s]")
    plt.ylabel("Quaternion")
    plt.show()


plot_orientation_vs_time("Orientation of last node of rod 1", pp_list_rod1["time"],
                         np.array(pp_list_rod1["director"])[..., -1])
plot_orientation_vs_time("Orientation of last node of rod 2", pp_list_rod2["time"],
                         np.array(pp_list_rod2["director"])[..., -1])
```

## Environment
- Python=3.10.5, pip=22.1.2 
- Numpy/numba version: 22.1.2
- PyElastica version: 0.2.3

## Expected behavior
We expect the FixedJoint to constrain all 6 DoF between the two rods: e.g. no translations or rotations. This should also include twisting / torsional movements in straight rod configuration.

## Screenshots
**Orientation of last node of rod 1 for torsion torque applied to rod 2:** 

Here, we would expect the rod 1 to rotate a bit around its z-axis to counteract the torsion torque applied at it's tip at the FixedJoint. However, it doesn't move at all.
fixed_joint_torsion_rod1

**Orientation of last node of rod 2 for torsion torque applied to rod 2:** 

Here, we would expect the rod 2 to rotate a bit around its z-axis to counteract the torsion torque. However, it is clearly not constrained at its base and just freely rotates.
fixed_joint_torsion_rod2

**Orientation of last node of rod 1 for bending torque applied to rod 2:** 

Here, we would expect the rod 1 to rotate a bit around its y-axis to counteract the torsion torque applied at it's tip at the FixedJoint.  This is exactly what happens and after a few seconds the elastic systems reaches its steady state.
fixed_joint_bending_rod1

**Orientation of last node of rod 12for bending torque applied to rod 2:** 

Here, we would expect the rod 2 to rotate a bit around its y-axis to counteract the torsion torque applied.  This is exactly what happens and after a few seconds the elastic systems reaches its steady state.
fixed_joint_bending_rod2

mstoelzle:
Fixes #131 for the FixedJoint class:
- Now any arbitrary torques applied to rods or rigid bodies connected by a FixedJoint are compensated (e.g. more general implementation)
- Rotation frame deviation is computed using rotation vectors
- Rotational spring-damper system is used to restrict the rotational mode
- **Spring:** spring constant `kt` scales the rotational frame deviation between systems one and two
- **Damper:**  damping constant `nut` scales the difference between angular velocities of system one and two
- Introduce functionality of static rotation offset between systems one and two: records the initial relative rotation between the two systems and enforces this static rotation throughout the entire simulation. If `use_static_rotation==False`, the static rotation is set to an identity matrix and restoring torques are applied to fully align the local frames of both systems at the join.
- Example `examples/JointCases/fixed_joint_torsion.py` illustrates how this implementation generates torsional / twisting torques to prevent torsional deviations
- Use common diagnostic callback for all joint cases examples
The attached plots demonstrates the results of the `examples/JointCases/fixed_joint_torsion.py` example:
Orientation_of_last_node_of_rod_1
Orientation_of_last_node_of_rod_2


mstoelzle:
@xzhan139 What is the reason that the damping force for the `SphericalJoint` constraints is projected onto the normalized distance error? 

```py
relative_velocity = (
            rod_two.velocity_collection[..., index_two]
            - rod_one.velocity_collection[..., index_one]
        )
        normal_relative_velocity = (
            np.dot(relative_velocity, normalized_end_distance_vector)
            * normalized_end_distance_vector
        )
        damping_force = -self.nu * normal_relative_velocity
```

When adding damping to the torque compensation for the `FixedJoint`, do we need this projection as well?

mstoelzle:
> @mstoelzle after discussing your question with @xzhan139, the conclusion is that there is no need to project it onto the normal vector. You can use directly the relative velocity field. In reality, contact damping is optional, since the internal dissipation of the rod helps in stabilizing the simulations.
> 
> > @xzhan139 What is the reason that the damping force for the `SphericalJoint` constraints is projected onto the normalized distance error?
> > ```python
> > relative_velocity = (
> >             rod_two.velocity_collection[..., index_two]
> >             - rod_one.velocity_collection[..., index_one]
> >         )
> >         normal_relative_velocity = (
> >             np.dot(relative_velocity, normalized_end_distance_vector)
> >             * normalized_end_distance_vector
> >         )
> >         damping_force = -self.nu * normal_relative_velocity
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When adding damping to the torque compensation for the `FixedJoint`, do we need this projection as well?
> 
> After you have done all necessary changes to the PR, please let us know or remove the WIP keyword from the PR name to open it for review.

@bhosale2 This PR is now ready for review

examples/JointCases/joint_cases_callback.py
```diff
@@ -0,0 +1,22 @@
+from elastica.callback_functions import CallBackBaseClass
+
+
+class JointCasesCallback(CallBackBaseClass):
```
skim0119, line 4: You can make alias `JointCasesCallback = MyCallback` instead to reduce code duplication. We want to minimize extra modules within the example cases.

elastica/joint.py
```diff
@@ -222,51 +228,65 @@ def __init__(self, k, nu, kt):
             Damping coefficient of the joint.
         kt: float
             Rotational stiffness coefficient of the joint.
+        nut: float = 0.
+            Rotational damping coefficient of the joint.
+        use_static_rotation: bool = True
+            If True, the initial rotation offset between the two systems is recorded
+            and enforced throughout the entire simulation.
```
skim0119, line 235:
```suggestion
            and enforced throughout the entire simulation. (default=True)
```

elastica/joint.py
```diff
@@ -222,51 +228,65 @@ def __init__(self, k, nu, kt):
...
+        # relative rotation matrix from system 1 to system 2: C_12 = C_1I @ C_I2 where I denotes the inertial frame
+        rel_rot = system_one_director @ system_two_director.T
+        # C_22* = C_21 @ C_12* where C_12* is the desired rotation between systems one and two
+        error_rot = rel_rot.T @ self.static_rotation
 
-        curr_destination = rod_two.position_collection[
-            ..., index_two + 1
-        ]  # second element of rod2
+        # compute rotation vectors based on C_22*
+        rot_vec = Rotation.from_matrix(error_rot).as_rotvec()
```
skim0119, line 270: You can use `_get_rotation_matrix` instead to be consistent, and for future `numba` wrapping.

elastica/joint.py
```diff
@@ -222,51 +228,65 @@ def __init__(self, k, nu, kt):
...
 
-        # Compute the restoring torque
-        forcedirection = -self.kt * (
-            curr_destination - tgt_destination
-        )  # force direction is between rod2 2nd element and rod1
-        torque = np.cross(link_direction, forcedirection)
+        # rotate rotation vector into inertial frame
+        rot_vec = system_two_director.T @ rot_vec
```
skim0119, line 273:
```suggestion
        rot_vec_inertial_frame = system_two_director.T @ rot_vec
```

skim0119 (changes requested):
I only have minor comments. I think the implementation is not wrong but not so clear. @bhosale2 Please take a look.

I would suggest avoiding any branching, and make `FixedJoint` with rotational offset as a default behavior. @armantekinalp
If we need branching (`static_rotation`), we should have separate test for each cases.

examples/JointCases/joint_cases_callback.py
mstoelzle: Thank your suggestion. Fixed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/5b379506e3dda6c2344c945d85759955737108c2

elastica/joint.py
```diff
@@ -222,51 +228,65 @@ def __init__(self, k, nu, kt):
...
-            rod_one.director_collection[..., index_one] @ torque
-        )
-        rod_two.external_torques[..., index_two] += (
-            rod_two.director_collection[..., index_two] @ torque
-        )
+        # we compute the constraining torque using a rotational spring - damper system in the inertial frame
+        torque = self.kt * rot_vec
+
+        # add damping torque
+        if self.nut > 0.0:
```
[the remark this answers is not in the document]
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/d8d7be80bc56ff59f998dfa7e3cbe05dba870591

elastica/joint.py
```diff
@@ -222,51 +228,65 @@ def __init__(self, k, nu, kt):
...
+        # collect directors of systems one and two
+        # note that systems can be either rods or rigid bodies
+        system_one_director = system_one.director_collection[..., index_one]
+        system_two_director = system_two.director_collection[..., index_two]
 
-        # To constrain the orientation of link two, the second node of link two should align with
-        # the direction of link one. Thus, we compute the desired position of the second node of link two
-        # as check1, and the current position of the second node of link two as check2. Check1 and check2
-        # should overlap.
+        if self.static_rotation is None:
```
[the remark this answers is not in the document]
mstoelzle: I require access to the directors of both systems to initialize `self.static_rotation`. In the `__init__`, I do not have access to the systems. Do you have an idea how to do this?

elastica/joint.py
```diff
@@ -222,51 +228,65 @@ def __init__(self, k, nu, kt):
...
+            # this if clause should be active during the first timestep for the case use_static_rotation==True
+            self.static_rotation = system_one_director @ system_two_director.T
 
-        tgt_destination = (
-            rod_one.position_collection[..., index_one]
-            + rod_two.rest_lengths[index_two] * rod_one.tangents[..., index_one]
-        )  # dl of rod 2 can be different than rod 1 so use rest length of rod 2
+        # relative rotation matrix from system 1 to system 2: C_12 = C_1I @ C_I2 where I denotes the inertial frame
+        rel_rot = system_one_director @ system_two_director.T
+        # C_22* = C_21 @ C_12* where C_12* is the desired rotation between systems one and two
```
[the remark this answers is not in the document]
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/00ecf2f9f0b6ec6cbbe68af802052ed6e7edd932

elastica/joint.py
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/00ecf2f9f0b6ec6cbbe68af802052ed6e7edd932

elastica/joint.py
```diff
@@ -222,51 +228,65 @@ def __init__(self, k, nu, kt):
             Damping coefficient of the joint.
         kt: float
             Rotational stiffness coefficient of the joint.
+        nut: float = 0.
+            Rotational damping coefficient of the joint.
+        use_static_rotation: bool = True
```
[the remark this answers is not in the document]
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/00ecf2f9f0b6ec6cbbe68af802052ed6e7edd932

elastica/joint.py
```diff
@@ -209,9 +210,14 @@ class FixedJoint(FreeJoint):
             Damping coefficient of the joint.
         kt: float
             Rotational stiffness coefficient of the joint.
+        nut: float
+            Rotational damping coefficient of the joint.
+        static_rotation: np.array
+            Static 3x3 rotation matrix from system one to system two.
+            Instead of aligning the directors of both systems directly, a desired rotational offset C_12* is enforced.
```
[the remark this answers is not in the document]
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/00ecf2f9f0b6ec6cbbe68af802052ed6e7edd932

elastica/joint.py
mstoelzle: @skim0119 `_get_rotation_matrix` is not suitable as it computes the rotation matrix. I need to compute the rotation vector based on a rotation matrix. 

Do you think it is advantageous to implement this with numba? I am asking, because we do not need any batched processing in this instance, but only need to compute the rotation vector for one rotation matrix.

skim0119: We have `_inv_rotate` for Rodrigues formula as well. You can use this.

We probably don't have much speed up with `numba` in this case, but we are trying to keep the code consistent for later. We are constantly checking if there are new traits available for optimization (`numba` for past example), and we want to keep dependencies as consistent as possible, especially for the critical kernels. It is why we are a little hesitant in utilizing high-level packages, especially during the stepping.

bhosale2: Maybe `inv_rotate` is what you may need @mstoelzle?

elastica/joint.py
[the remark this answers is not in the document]
bhosale2: The name `static_rotation` implies it's a static quantity set at initialization time. @mstoelzle is it possible to compute these at an initial time in say one rod's local frame and then reuse this fixed matrix? As for the access to the rods, maybe @skim0119 can clarify a cleaner route.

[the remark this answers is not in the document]
skim0119: We will support accessing systems (`rod1` and `rod2`) during the initialization, so please make a `TODO` label here for future refactoring. It is not on the immediate plan, so I would recommend computing `offset` orientation outside the function and passing it as an initialization parameter. 

Below is an example of `parallel connection`, in which we have a similar issue. We provide a separate method to compute initial offset vectors and pass those vectors as a connection argument.
[parallel connection](https://github.com/GazzolaLab/PyElastica/blob/master/elastica/experimental/connection_contact_joint/parallel_connection.py)

elastica/joint.py
bhosale2, line 217:
```suggestion
            Rest 3x3 rotation matrix from system one to system two at the connected elements.
            Instead of aligning the directors of both systems directly, a desired rest rotational matrix labeled C_12 is enforced.
```

elastica/joint.py
```diff
@@ -209,9 +210,14 @@ class FixedJoint(FreeJoint):
             Damping coefficient of the joint.
         kt: float
             Rotational stiffness coefficient of the joint.
+        nut: float
+            Rotational damping coefficient of the joint.
+        static_rotation: np.array
```
bhosale2, line 215: I feel `static_rotation` is less intuitive and maybe `rest_rotation_matrix` is more intuitive? since it is the rest orientation of the 2 connected elements.

elastica/joint.py
```diff
@@ -222,51 +228,68 @@ def __init__(self, k, nu, kt):
             Damping coefficient of the joint.
         kt: float
             Rotational stiffness coefficient of the joint.
+        nut: float = 0.
+            Rotational damping coefficient of the joint.
+        use_static_rotation: bool
+            If True, the initial rotation offset between the two systems is recorded
+            and enforced throughout the entire simulation. (default=True)
```
bhosale2, line 235:
```suggestion
        is_rest_joint_state: bool
            If True, the initial rotation angle between the two systems is recorded
            and enforced throughout the entire simulation. (default=True)
```

elastica/joint.py
```diff
@@ -209,9 +210,14 @@ class FixedJoint(FreeJoint):
...
             Rotational stiffness coefficient of the joint.
+        nut: float
+            Rotational damping coefficient of the joint.
+        static_rotation: np.array
+            Static 3x3 rotation matrix from system one to system two.
+            Instead of aligning the directors of both systems directly, a desired rotational offset C_12* is enforced.
     """
 
-    def __init__(self, k, nu, kt):
+    def __init__(self, k, nu, kt, nut=0.0, use_static_rotation=True):
```
bhosale2, line 220:
```suggestion
    def __init__(self, k, nu, kt, nut=0.0, is_rest_joint_state=True):
```

elastica/joint.py
```diff
@@ -222,51 +228,68 @@ def __init__(self, k, nu, kt):
...
         # stiffness of the restoring constraint -- tuned empirically
         self.kt = kt
+        self.nut = nut
+
+        # if a static rotation offset between the two systems should not be enforced,
+        # we set the relative rotation to the identity matrix. Otherwise
+        if use_static_rotation:
+            self.static_rotation = None
+        else:
+            self.static_rotation = np.eye(3)
```
bhosale2, line 248:
```suggestion
        # if the current rotation between the 2 rods is not the rest state,
        # then we set the rest rotation to the identity matrix.
        if not is_rest_joint_state:
            self.rest_rotation_matrix = np.eye(3)
```

elastica/joint.py
```diff
@@ -222,51 +228,68 @@ def __init__(self, k, nu, kt):
...
-        # also NOTE: - rod two is fixed at first element
-        link_direction = (
-            rod_two.position_collection[..., index_two + 1]
-            - rod_two.position_collection[..., index_two]
+    def apply_torques(self, system_one, index_one, system_two, index_two):
+        # collect directors of systems one and two
+        # note that systems can be either rods or rigid bodies
+        system_one_director = system_one.director_collection[..., index_one]
+        system_two_director = system_two.director_collection[..., index_two]
+
```
bhosale2, line 259: with the suggested changes above this line, you won't need a branching here. Just make the code in the branch below as default.

elastica/joint.py
```diff
@@ -222,51 +228,68 @@ def __init__(self, k, nu, kt):
...
+        if self.static_rotation is None:
+            # this if clause should be active during the first timestep for the case use_static_rotation==True
+            self.static_rotation = system_one_director @ system_two_director.T
```
bhosale2, line 262: won't be needed if these things are set in initializer.

elastica/joint.py
```diff
@@ -222,51 +228,68 @@ def __init__(self, k, nu, kt):
...
+        # rel_rot: C_12 = C_1I @ C_I2
+        # C_12 is relative rotation matrix from system 1 to system 2
+        # C_1I is the rotation from system 1 to the inertial frame (i.e. the world frame)
+        # C_I2 is the rotation from the inertial frame to system 2 frame (inverse of system_two_director)
+        rel_rot = system_one_director @ system_two_director.T
+        # error_rot: C_22* = C_21 @ C_12*
+        # C_22* is rotation matrix from current orientation of system 2 to desired orientation of system 2
+        # C_21 is the inverse of C_12, which describes the relative (current) rotation from system 1 to system 2
+        # C_12* is the desired rotation between systems one and two, which is saved in the static_rotation attribute
+        error_rot = rel_rot.T @ self.static_rotation
```
bhosale2, line 273:
```suggestion
        dev_rot = rel_rot.T @ self.rest_rotation_matrix
```

elastica/joint.py
```diff
@@ -222,51 +228,68 @@ def __init__(self, k, nu, kt):
...
+        rot_vec = Rotation.from_matrix(error_rot).as_rotvec()
+
+        # rotate rotation vector into inertial frame
+        rot_vec_inertial_frame = system_two_director.T @ rot_vec
+
+        # error in rotation velocity between system 1 and system 2
+        error_omega = (
+            system_one.omega_collection[..., index_one]
+            - system_two.omega_collection[..., index_two]
         )
```
bhosale2, line 285:
```suggestion
        # deviation in rotation velocity between system 1 and system 2 
        # first convert to lab frame, then take differences
        dev_omega = (
            system_one_director.T @ system_one.omega_collection[..., index_one]
            - system_two_director.T @ system_two.omega_collection[..., index_two]
        )
```

elastica/joint.py
```diff
@@ -222,51 +228,68 @@ def __init__(self, k, nu, kt):
...
-        tgt_destination = (
-            rod_one.position_collection[..., index_one]
-            + rod_two.rest_lengths[index_two] * rod_one.tangents[..., index_one]
-        )  # dl of rod 2 can be different than rod 1 so use rest length of rod 2
-
-        curr_destination = rod_two.position_collection[
-            ..., index_two + 1
-        ]  # second element of rod2
+        # we compute the constraining torque using a rotational spring - damper system in the inertial frame
+        torque = self.kt * rot_vec_inertial_frame - self.nut * error_omega
```
bhosale2, line 288:
```suggestion
        torque = self.kt * rot_vec_inertial_frame - self.nut * dev_omega
```

bhosale2 (changes requested):
good job overall @mstoelzle, some critical comments to be addressed.

elastica/joint.py
mstoelzle: Resolved in https://github.com/GazzolaLab/PyElastica/pull/135/commits/3b7a1fdbebe2b8fa3c9b1a6430a25defed70e746

elastica/joint.py
mstoelzle: Resolved in https://github.com/GazzolaLab/PyElastica/pull/135/commits/9a5cf1cb90ebd7fb6a215d799eb27c5d0bd1c5f0

elastica/joint.py
mstoelzle: Resolved in https://github.com/GazzolaLab/PyElastica/pull/135/commits/9a5cf1cb90ebd7fb6a215d799eb27c5d0bd1c5f0

elastica/joint.py
mstoelzle: Resolved in https://github.com/GazzolaLab/PyElastica/pull/135/commits/9a5cf1cb90ebd7fb6a215d799eb27c5d0bd1c5f0

elastica/joint.py
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/cb84464402707586dbc511df29a4156d42a7bdc0

elastica/joint.py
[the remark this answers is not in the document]
mstoelzle: Addressed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/cb84464402707586dbc511df29a4156d42a7bdc0

elastica/joint.py
mstoelzle: Removed branching in https://github.com/GazzolaLab/PyElastica/pull/135/commits/8fd8976465a320976915979d24d84020641f643e

elastica/joint.py
mstoelzle: As `rest_rotation_matrix` is now passed as a parameter to the constructor of RigidJoint, this code is not necessary anymore and was removed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/cb84464402707586dbc511df29a4156d42a7bdc0

armantekinalp (changes requested):
Good job @mstoelzle minor changes.

elastica/joint.py
```diff
@@ -209,9 +210,15 @@ class FixedJoint(FreeJoint):
...
+        nut: float
+            Rotational damping coefficient of the joint.
+        rest_rotation_matrix: np.array
+            Rest 3x3 rotation matrix from system one to system two at the connected elements.
+            Instead of aligning the directors of both systems directly, a desired rest rotational matrix labeled C_12*
+            is enforced.
     """
 
-    def __init__(self, k, nu, kt):
+    def __init__(self, k, nu, kt, nut=0.0, rest_rotation_matrix=np.eye(3)):
```
skim0119, line 221: Avoid using numpy array as a default argument.

skim0119 (changes requested):

bhosale2 (approved):
LGTM. @skim0119 and @armantekinalp will follow up.

armantekinalp:
@mstoelzle some CI tests are failing can you merge changes from `update-0.3.0` to your branch.

mstoelzle:
> @mstoelzle some CI tests are failing can you merge changes from `update-0.3.0` to your branch. 

Thanks. I will do that. There is also still some test issues I want to have a look at.

elastica/joint.py
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/91d18173b492499cd574b0d13ca9b7a94dace42c

examples/JointCases/joint_cases_callback.py
skim0119: Please remove this file, I don't think it is used.

elastica/joint.py
```diff
@@ -222,51 +229,96 @@ def __init__(self, k, nu, kt):
...
+        # C_I2 is the rotation from the inertial frame to system 2 frame (inverse of system_two_director)
+        rel_rot = system_one_director @ system_two_director.T
+        # error_rot: C_22* = C_21 @ C_12*
+        # C_22* is rotation matrix from current orientation of system 2 to desired orientation of system 2
+        # C_21 is the inverse of C_12, which describes the relative (current) rotation from system 1 to system 2
+        # C_12* is the desired rotation between systems one and two, which is saved in the static_rotation attribute
+        dev_rot = rel_rot.T @ self.rest_rotation_matrix
+
+        # compute rotation vectors based on C_22*
+        rot_vec = Rotation.from_matrix(dev_rot).as_rotvec()
```
skim0119, line 278: Please use available tools in `_rotations.py`

tests/test_joint.py
```diff
@@ -309,7 +317,7 @@ def test_fixedjoint():
     dampingforce = nu * normal_relative_vel * distance / end_distance
     contactforce = elasticforce - dampingforce
 
-    fxjt = FixedJoint(k, nu, kt)
+    fxjt = FixedJoint(k, nu, kt, nut, rest_rotation_matrix=np.eye(3))
```
skim0119, line 320: Can we add more test cases with the non-initially-aligned scenario?

elastica/joint.py
```diff
@@ -222,51 +229,96 @@ def __init__(self, k, nu, kt):
...
-        rod_one.external_torques[..., index_one] -= (
-            rod_one.director_collection[..., index_one] @ torque
-        )
-        rod_two.external_torques[..., index_two] += (
-            rod_two.director_collection[..., index_two] @ torque
-        )
+def get_relative_rotation_two_systems(system_one, index_one, system_two, index_two):
+    """
+    Compute the relative rotation matrix C_12 between system one and system two at the specified elements.
+
```
skim0119, line 301:
```suggestion

    Examples
    ----------
    
        >>> init_rot = get_relative_rotation_two_systems(...)
        >>> ...
        
    See Also
    ---------
    FixedJoint 

```

Just because this function is a free function, can you provide an example usage script?

skim0119 (changes requested):
It is probably the last set of changes. After you fix them, I'll run the testings and `JointCases` code, and approve.

examples/JointCases/joint_cases_callback.py
mstoelzle: Addressed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/6d360f6352393a911ce0aa7a4425defd1fa5dee4

elastica/joint.py
```diff
@@ -209,9 +210,15 @@ class FixedJoint(FreeJoint):
             Damping coefficient of the joint.
         kt: float
             Rotational stiffness coefficient of the joint.
+        nut: float
+            Rotational damping coefficient of the joint.
+        rest_rotation_matrix: np.array
```
armantekinalp, line 215: minor change.
```suggestion
        rest_rotation_matrix: np.array
            2D (3,3) array containing data with 'float' type.
```

tests/test_joint.py
armantekinalp: I agree with @skim0119 . New `fixed joint` class is capable of more than connecting two aligned systems. Maybe add one more test for connecting two rods that are perpendicular to each other.

elastica/joint.py
mstoelzle: @skim0119 I am confused about the implementation of `_inv_rotate`. Could you please expand a bit what exactly it does? 

Particularly, I have a question concerning the shape of the return array. I would expect that similarly to other functions for example in `_linalg.py`, the functions expects inputs in batch format (with the `n` dimension last). However, the dimension of the last dim is calculated as `n-1` in line 97. This is also inconsistent with the docstring, which says that the return has a shape of `(dim, n)`. But in reality, it is `(dim, n-1)`.
```py
@njit(cache=True)
def _inv_rotate(director_collection):
    """
    Calculated rate of change using Rodrigues' formula

    Parameters
    ----------
    director_collection : The collection of frames/directors at every element,
    numpy.ndarray of shape (dim, dim, n)

    Returns
    -------
    vector_collection : The collection of axes around which the body rotates
    numpy.ndarray of shape (dim, n)

    Note
    ----
    TODO: Benchmark missing

    """
    blocksize = director_collection.shape[2] - 1
    vector_collection = np.empty((3, blocksize))
```

mstoelzle: @skim0119 Also, when I compare the outputs of Scipy `Rotation.from_matrix(dev_rot).as_rotvec()` and `_inv_rotate(dev_rot.reshape(3, 3, 1)).squeeze()` for the same rotation matrix `dev_rot.reshape `, the outputs are quite different:
```
dev_rot:  [[ 1.00000000e+00 -2.98624524e-07 -1.39620560e-10]
 [ 2.98624524e-07  1.00000000e+00  2.75057444e-07]
 [ 1.39537193e-10 -2.75057444e-07  1.00000000e+00]]
theta 4.060023871629602e-07 n [-6.77477405e-01 -3.43788316e-04  7.35524059e-01]
Scipy Rotation.from_matrix(dev_rot).as_rotvec(): [-2.75057444e-07 -1.39578877e-10  2.98624524e-07]
_inv_rotate(dev_rot.reshape(3, 3, 1)).squeeze(): [-1.20919958e+00 -1.05479264e-06 -1.20919958e+00]

```

bhosale2: Hi, @mstoelzle good catch regarding the documentation @armantekinalp take note, as you see `inv_rotate` actually takes in a collection of n directors and returns n - 1 rotation vectors, where the vectors correspond to the relative rotation vectors between consecutive directors in the collection. A possible way of using inv_rotate to get the rotation vector in this scenario is:
```
_inv_rotate(np.dstack([rel_rot, self.rest_rotation_matrix]))
```
Check once implemented if the signs are consistent.

elastica/joint.py
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/135/commits/035e0ed52f9408a67ccadc1bb7db06b766b2a584

elastica/joint.py
mstoelzle: @bhosale2 @armantekinalp @skim0119 Yes, I think it would be a good idea to improve the documentation on `_inv_rotate` and explaining its purpose, input & output a bit better. Possibly, it would also be nice to change the name of the method to something more intuitive.

tests/test_joint.py
mstoelzle: Thanks. I addressed this in https://github.com/GazzolaLab/PyElastica/pull/135/commits/d84f5f1234aad142f732c15908ba7e40dd2eb466

mstoelzle:
Thanks everybody for your feedback. I have addressed all the requests for changes (in case I didn't miss anything) and this PR should be ready for merging now.

armantekinalp (approved):
LGTM, nice work @mstoelzle

skim0119 (approved):

PR #138, opened 2022-07-15
Remove _elastica_numba folder while keeping _elastica_numba.py

### LINKED ISSUE #136: Removing _elastica_numba folder and add _elastica_numba.py for user notification
Currently `numba` has become a default feature in `pyelastica` and the separate `_elastica_numba` module is deprecated since the past few version releases. The presence of that folder in future releases may create confusion as to if `numba` needs to be turned on or off in the simulator. As such, after discussion with @armantekinalp, we feel it may be time to remove that module altogether.

bhosale2:
Fixes #136.


skim0119 (approved):
LGTM

PR #147, opened 2022-07-19
Fix #142 by adjusting data structure of `fixed_positions` and `fixed_directors`

### LINKED ISSUE #142: Inconsistent docstring in `FixedConstraint` class
**Describe the bug**
The docstrings of the `nb_constraint_rotational_values`, `nb_constrain_translational_values` methods are not consistent with the implementation. The documentation mentions that both the `position_collection` and `fixed_position_collection` are in the shape `(dim, blocksize)`. But if you look at the code, `fixed_position_collection` is (probably) in the shape `(blocksize , dim)`: 
```py
@staticmethod
    @njit(cache=True)
    def nb_constraint_rotational_values(
        director_collection, fixed_director_collection, indices
    ) -> None:
        """
        Computes constrain values in numba njit decorator
        Parameters
        ----------
        director_collection : numpy.ndarray
            3D (dim, dim, blocksize) array containing data with `float` type.
        fixed_director_collection : numpy.ndarray
            3D (dim, dim, blocksize) array containing data with `float` type.
        indices : numpy.ndarray
            1D array containing the index of constraining nodes
        """
        block_size = indices.size
        for i in range(block_size):
            k = indices[i]
            director_collection[..., k] = fixed_director_collection[i, ...]

    @staticmethod
    @njit(cache=True)
    def nb_constrain_translational_values(
        position_collection, fixed_position_collection, indices
    ) -> None:
        """
        Computes constrain values in numba njit decorator
        Parameters
        ----------
        position_collection : numpy.ndarray
            2D (dim, blocksize) array containing data with `float` type.
        fixed_position_collection : numpy.ndarray
            2D (dim, blocksize) array containing data with `float` type.
        indices : numpy.ndarray
            1D array containing the index of constraining nodes
        """
        block_size = indices.size
        for i in range(block_size):
            k = indices[i]
            position_collection[0, k] = fixed_position_collection[i, 0]
            position_collection[1, k] = fixed_position_collection[i, 1]
            position_collection[2, k] = fixed_position_collection[i, 2]
```
The same issue exists analogue for `nb_constraint_rotational_values `

mstoelzle:
Fixes #142 by implementing the following changes:

- Tranpose `fixed_positions` and `fixed_directors` after initialization to mirror the standard data structure in `pyelastica`, which is batch-last as in `(dim, batch)` or `(dim, dim,batch)`
- Simplify code by using `...` for slicing of vector entries instead of listing each dim separately

### DIFF
--- a/elastica/boundary_conditions.py
+++ b/elastica/boundary_conditions.py
@@ -308,8 +308,20 @@ class FixedConstraint(ConstraintBase):
             else:
                 # TODO: This part is prone to error.
                 break
-        self.fixed_positions = np.array(pos)
-        self.fixed_directors = np.array(dir)
+
+        if len(pos) > 0:
+            # transpose from (blocksize, dim) to (dim, blocksize)
+            self.fixed_positions = np.array(pos).transpose((1, 0))
+        else:
+            # initialize empty fixed positions
+            self.fixed_positions = np.zeros((3, 0))
+
+        if len(dir) > 0:
+            # transpose from (blocksize, dim, dim) to (dim, dim, blocksize)
+            self.fixed_directors = np.array(dir).transpose((1, 2, 0))
+        else:
+            # initialize empty fixed directors
+            self.fixed_directors = np.zeros((3, 3, 0))
 
     def constrain_values(
         self, rod: Union[Type[RodBase], Type[RigidBodyBase]], time: float
@@ -362,7 +374,7 @@ class FixedConstraint(ConstraintBase):
         block_size = indices.size
         for i in range(block_size):
             k = indices[i]
-            director_collection[..., k] = fixed_director_collection[i, ...]
+            director_collection[..., k] = fixed_director_collection[..., i]
 
     @staticmethod
     @njit(cache=True)
@@ -385,9 +397,7 @@ class FixedConstraint(ConstraintBase):
         block_size = indices.size
         for i in range(block_size):
             k = indices[i]
-            position_collection[0, k] = fixed_position_collection[i, 0]
-            position_collection[1, k] = fixed_position_collection[i, 1]
-            position_collection[2, k] = fixed_position_collection[i, 2]
+            position_collection[..., k] = fixed_position_collection[..., i]
 
     @staticmethod
     @njit(cache=True)
@@ -406,9 +416,7 @@ class FixedConstraint(ConstraintBase):
         block_size = indices.size
         for i in range(block_size):
             k = indices[i]
-            velocity_collection[0, k] = 0.0
-            velocity_collection[1, k] = 0.0
-            velocity_collection[2, k] = 0.0
+            velocity_collection[..., k] = 0.0
 
     @staticmethod
     @njit(cache=True)
@@ -427,9 +435,7 @@ class FixedConstraint(ConstraintBase):
         block_size = indices.size
         for i in range(block_size):
             k = indices[i]
-            omega_collection[0, k] = 0.0
-            omega_collection[1, k] = 0.0
-            omega_collection[2, k] = 0.0
+            omega_collection[..., k] = 0.0
 
 
 class HelicalBucklingBC(ConstraintBase):


elastica/boundary_conditions.py
```diff
@@ -308,8 +308,20 @@ def __init__(self, *fixed_data, **kwargs):
...
                 break
-        self.fixed_positions = np.array(pos)
-        self.fixed_directors = np.array(dir)
+
+        if len(pos) > 0:
+            # transpose from (blocksize, dim) to (dim, blocksize)
+            self.fixed_positions = np.array(pos).transpose((1, 0))
+        else:
+            # initialize empty fixed positions
+            self.fixed_positions = np.zeros((3, 0))
```
bhosale2, line 317:
```suggestion
```
not needed since a non-empty list is always generated in the wrapper.

elastica/boundary_conditions.py
```diff
@@ -308,8 +308,20 @@ def __init__(self, *fixed_data, **kwargs):
...
+
+        if len(dir) > 0:
+            # transpose from (blocksize, dim, dim) to (dim, dim, blocksize)
+            self.fixed_directors = np.array(dir).transpose((1, 2, 0))
+        else:
+            # initialize empty fixed directors
+            self.fixed_directors = np.zeros((3, 3, 0))
```
bhosale2, line 324:
```suggestion
```

bhosale2 (changes requested):
LGTM mostly, super minor changes.

elastica/boundary_conditions.py
mstoelzle: I am not quite sure where the non-empty list is generated in the wrapper, but the test also runs through successfully without this else statement. So I removed it in https://github.com/GazzolaLab/PyElastica/pull/147/commits/c57b88d9cec9e8590f9305365acc6fe59026b022

elastica/boundary_conditions.py
mstoelzle: Removed in https://github.com/GazzolaLab/PyElastica/pull/147/commits/c57b88d9cec9e8590f9305365acc6fe59026b022

armantekinalp (approved):
LGTM good catch!

bhosale2 (approved):
LGTM.

elastica/boundary_conditions.py
```diff
@@ -406,9 +410,7 @@ def nb_constrain_translational_rates(velocity_collection, indices) -> None:
         block_size = indices.size
         for i in range(block_size):
             k = indices[i]
-            velocity_collection[0, k] = 0.0
-            velocity_collection[1, k] = 0.0
-            velocity_collection[2, k] = 0.0
+            velocity_collection[..., k] = 0.0
```
skim0119, line 413: Can we revert this change back? It includes extra instructor steps to broadcast and is typically faster to represent by unrolling vector. The above `position_collection[...,k] = fixed_position_collection[...,i]` is still okay because the number of instructions in the copying vector is probably the same.

It is a bit controversial topic since this step probably affects minimally on overall walltime, and it probably doesn't affect much. However, most of the old kernels we have are often tested multiple times with multiple implementations, and we don't want to change them for code simplification. Timing is a bit sensitive topic in our project, so please understand.

## code

``` py
import numpy as np
from numba import njit

N = 100
vec = np.random.random((3, N))

# --------------- Test Set Zero -------------- #
def f1(a):
    block_size = a.shape[1]
    for i in range(block_size):
        a[0,i] = 0.0
        a[1,i] = 0.0
        a[2,i] = 0.0
    
def f2(a):
    block_size = a.shape[1]
    for i in range(block_size):
        a[...,i] = 0.0
```

## result (N=100)

```py
In [1]: run test.py

In [2]: %timeit f1(vec)
27.3 µs ± 1.08 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [3]: %timeit f2(vec)
37.6 µs ± 722 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

## result (N=10000)

```py
In [1]: run test.py

In [2]: %timeit f1(vec)
2.8 ms ± 59.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [3]: %timeit f2(vec)
4.17 ms ± 121 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

elastica/boundary_conditions.py
```diff
@@ -427,9 +429,7 @@ def nb_constrain_rotational_rates(omega_collection, indices) -> None:
         block_size = indices.size
         for i in range(block_size):
             k = indices[i]
-            omega_collection[0, k] = 0.0
-            omega_collection[1, k] = 0.0
-            omega_collection[2, k] = 0.0
+            omega_collection[..., k] = 0.0
```
skim0119, line 432: ditto ^

skim0119 (changes requested):
LGTM. One change request

elastica/boundary_conditions.py
bhosale2: @skim0119 can you post the timings with `njit` operator applied?

skim0119:
```py
In [1]: run test.py

In [2]: f1(vec); f2(vec); # Cache, First run

In [3]: %timeit -r 10 f1(vec)
261 ns ± 1.24 ns per loop (mean ± std. dev. of 10 runs, 1000000 loops each)

In [4]: %timeit -r 10 f2(vec)
515 ns ± 9.93 ns per loop (mean ± std. dev. of 10 runs, 1000000 loops each)
```

bhosale2: Ok, I feel 0.5 us is hardly gonna affect the rod timestep timing which for 100 elements is roughly 50-60 us, but I agree with @skim0119 regarding reverting the change since it does not lead to that much of a code simplification.

skim0119: @bhosale2 `numba` helps compile the code, but it probably doesn't optimize instruction itself, further than what LLVM does. For LLVM, we manually need to add `#pragma unroll` to specify SIMD vectorization, so I guess similar logic applies here.

elastica/boundary_conditions.py
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/147/commits/9a95c4389de20fce48cc9771faeb714f4694607c

elastica/boundary_conditions.py
mstoelzle: Thanks for the explanation. That is indeed quite interesting. I reverted the changes in https://github.com/GazzolaLab/PyElastica/pull/147/commits/9a95c4389de20fce48cc9771faeb714f4694607c

skim0119 (approved):

PR #143, opened 2022-07-18
Address #148 by adding `ConfigurableFixedConstraint` boundary condition class

mstoelzle:
## Motivation
Currently, there exists only the option to either have a) no boundary conditions or b) fully fixed boundary conditions. The only functionality in between is that for example translations are allowed, but rotations are constrained and vice-versa. However in practice, there are many situations where we would like to allow certain translations (for example along a plane) or rotations (for example allow yawing of the end of the rod), but constrain all other DoF. The user of the library would have to implement this functionality all themselves inheriting from the `FreeBC` class. In many cases, such an implementation is not straight-forward. Thus, the library should provide the user the option to configure the BC themselves in a simple fashion. Although it obviously will never cover all the edge-cases a specific project might require, it will give the user an idea / hint / guidance on how to implement a similar solution.

## Provided functionality
This PR add the following functionality through the `ConfigurableFixedConstraint` class:

- Using boolean arrays `translational_constraint_selector` and `rotational_constraint_selector`, the user can specify which Degrees of Freedom to allow / constrain at the specified nodes.
- These boolean arrays have to be specified in the inertial frame.
- The `ConfigurableFixedConstraint` will now only re-set the positions, velocities, directors and omegas to the saved, fixed values for the constrained dimensions. The other DoF are allowed to freely move / rotate.
- We can simplify the codebase by now having `FixedConstraint` be a child class of `ConfigurableFixedConstraint`. `FixedConstraint` will still constrain all dimension, e.g. `translational_constraint_selector=np.array([True, True, True])` and `rotational_constraint_selector=np.array([True, True, True])`

## Examples for usage
Below examples motivate the usage of this new class.
```py
# How to fix all translational and rotational DoF except allowing twisting around z-axis in inertial frame:

simulator.constrain(rod).using(
    ConfigurableFixedConstraint,
    constrained_position_idx=(0,),
    constrained_director_idx=(0,),
    translational_constraint_selector=np.array([True, True, True]),
    rotational_constraint_selector=np.array([True, True, False]),
)

# How to allow the end of the rod to move in the x-y plane and allow all rotational DoF:

simulator.constrain(rod).using(
    ConfigurableFixedConstraint,
    constrained_position_idx=(-1,),
    translational_constraint_selector=np.array([True, True, False]),
)
```

I also added an example in `examples/BoundaryConditionCases/configurable_fixed_constraint.py`, which applies torsion to end tip of a rod. With a fully `FixedConstraint` the resulting orientation of the first node of the rod would look like this:

[...]

Issue #113, opened 2022-06-16
Separate joint module into separate connections and collision


bhosale2:
Currently, the implementation for connections and collision lies in `joint.py` which is a rather non-intuitive name for its contents. A more intuitive name could be `connections_and_collision.py`? @armantekinalp and @skim0119 suggestions?

skim0119:
I think `collision` should be separated. I agree that `connection could be more general name.

bhosale2:
Ok, then the consensus on this is:
1. Rename `joint.py` to `connections.py`
2. Move out collision classes into a separate `collision.py`

armantekinalp:
Yes, I think it is a good idea

Ali-7800:
Merging checklist
- [ ] New contact module and tests
- [ ] Updated rod-rod, rod-cylinder, self-contact classes, examples, and tests
- [ ] Surface class, surface subclass: plane, and tests
- [ ] Updated rod-plane, cylinder-plane classes, examples, and tests

Rahul-JOON:
@Ali-7800 added the New Contact module, surface base class, and tests via #281.

Ali-7800:
Added Updated rod-rod, rod-cylinder, self-contact classes, examples, and tests via #291

Ali-7800:
Added the new plane surface subclass.

Ali-7800:
Added updated rod-plane contact class

Ali-7800:
Updated rod-plane contact classes tests and refactored contact functions in interaction.py and joint.py

Ali-7800:
Updated rod-plane contact examples and added continuum snake with lifting wave example

Ali-7800:
Updated documentation for contact module, closing now as issue is resolved

PR #149, opened 2022-07-20
Fix #122 by enabling joints to connect rods and rigid bodies 

### LINKED ISSUE #122: Joint between Rod and RigidBody
**Describe the bug**
According to the [documentation](https://docs.cosseratrods.org/en/latest/api/connections.html#elastica.joint.FixedJoint), a numerous number of joints (and particularly the FixedJoint) should have compatibility for connecting a CosseratRod and a RigidBody (for example a Cylinder). However in practice, there are some bugs when trying to use this feature:

Error message for call `simulator.connect(rigid_body, rod, first_connect_idx=0, second_connect_idx=-1)`:
```
File "/Users/maximilianstoelzle/opt/miniconda3/envs/hsa-control/lib/python3.10/site-packages/elastica/wrappers/connections.py", line 99, in _call_connections
    connection.apply_torques(
  File "/Users/maximilianstoelzle/opt/miniconda3/envs/hsa-control/lib/python3.10/site-packages/elastica/joint.py", line 250, in apply_torques
    + rod_two.rest_lengths[index_two] * rod_one.tangents[..., index_one]
AttributeError: 'Cylinder' object has no attribute 'tangents'
```

Error message for call `simulator.connect(rod, rigid_body, first_connect_idx=0, second_connect_idx=0)`:
```
 File "/Users/maximilianstoelzle/opt/miniconda3/envs/hsa-control/lib/python3.10/site-packages/elastica/wrappers/connections.py", line 99, in _call_connections
    connection.apply_torques(
  File "/Users/maximilianstoelzle/opt/miniconda3/envs/hsa-control/lib/python3.10/site-packages/elastica/joint.py", line 250, in apply_torques
    + rod_two.rest_lengths[index_two] * rod_one.tangents[..., index_one]
AttributeError: 'Cylinder' object has no attribute 'rest_lengths'
```

Error message for call `simulator.connect(rod, rigid_body, first_connect_idx=-1, second_connect_idx=0)`:
```
 File "/Users/maximilianstoelzle/opt/miniconda3/envs/hsa-control/lib/python3.10/site-packages/elastica/wrappers/connections.py", line 99, in _call_connections
    connection.apply_torques(
  File "/Users/maximilianstoelzle/opt/miniconda3/envs/hsa-control/lib/python3.10/site-packages/elastica/joint.py", line 239, in apply_torques
    rod_two.position_collection[..., index_two + 1]
IndexError: index 1 is out of bounds for axis 1 with size 1
```

**To Reproduce**
```py
from elastica.joint import FixedJoint
from elastica.rod.cosserat_rod import CosseratRod
from elastica.rigidbody import Cylinder
rod = CosseratRod.straight_rod(
                n_elements=50,  # number of elements
                start=rod_base_position,  # Starting position of first node in rod
                direction=np.array([0.0, 1.0, 0.0]),  # Direction the rod extends
                normal=np.array([0.0, 0.0, 1.0]),  # normal vector of rod
                base_length=1.,  # original length of rod (m)
                base_radius=0.05,  # original radius of rod (m)
                density=1e3,  # density of rod (kg/m^3)
                nu=1e-3,  # Energy dissipation of rod
                youngs_modulus=youngs_modulus,
                shear_modulus=shear_modulus,
            )
rigid_body =  Cylinder(start=np.array[0., 0., 1.),
                            normal=np.array([0., 0., 1.]),
                            direction=np.array([1., 0., 0.]),
                            base_length=platform_thickness,
                            base_radius=platform_radius,
                            density=1.,  # TODO: replace with actual density of platform
                        )
simulator.connect(rigid_body, rod, first_connect_idx=0, second_connect_idx=-1).using(
                                    FixedJoint,
                                    k=1e8, kt=1e6, nu=1,
                                )
```

**Environment**
- Python=3.10.5, pip=22.1.2 
- Numpy/numba version: 22.1.2
- PyElastica version: 0.2.3

**Expected behavior**
I expect the be able to connect a RigidBody and a CosseratRod via a (Fixed)Joint, no matter which is referenced as `rod_1` or `rod_2` in the `connect` method.

**Screenshots**

**Additional context**

mstoelzle:
## Features and Changes
This issue fixes #122 by implementing the following changes and features:

1. Enables the connection of Cosserat rods and rigid bodies in any arbitrary combination for the following joint types: `FreeJoint`, `HingeJoint`, and `FixedJoint`
2. The `FreeJoint` constrains the translational movement of both systems at the specified connection point. As the point of joint connection doesn't necessarily coincide with the center of mass of the rigid body or the node of the rod, the user can use the `point_system_one` and `point_system_two` parameters to specify how the connection point relates to the node / CoM in the local frame of the system.
3. For the `HingeJoint`,  the `link_direction` of the second system is now extracted from the `director_collection` instead of subtracting node positions
4. The `FixedJoint` class now also accepts the `point_system_one` and `point_system_two` parameters and forwards them to the parent class `FreeJoint`

## Examples of Usage
How to connect two Cosserat rods together using a spherical joint with a gap of 0.01 m in between.

```py
simulator.connect(rod_one, rod_two).using(
    FreeJoint,
    k=1e4,
    nu=1,
    point_system_one=np.array([0.0, 0.0, 0.005]),
    point_system_two=np.array([0.0, 0.0, -0.005]),
)
```
    
How to connect the distal end of a CosseratRod with the base of a cylinder using a spherical joint.

```py
simulator.connect(rod, cylinder).using(
    FreeJoint,
    k=1e4,
    nu=1,
    point_system_two=np.array([0.0, 0.0, -cylinder.length / 2.]),
)
```

How to connect a cosserat rod with the base of a cylinder using a fixed joint, where the cylinder is rotated by 45 degrees around the y-axis.

```py
from scipy.spatial.transform import Rotation
simulator.connect(rod, cylinder).using(
      FixedJoint,
      k=1e5,
      nu=1e0,
      kt=1e3,
      nut=1e-3,
      point_system_two=np.array([0, 0, -cylinder.length / 2]),
      rest_rotation_matrix=Rotation.from_euler('y', np.pi / 4, degrees=False).as_matrix(),
)
```

## Results from `JointCases`
**Spherical joint:** 

System consisting of two rods and one cylinder connected with spherical joints. The sinusoidal force is applied to the second rod. Example can be run using:

```bash
python examples/JointCases/spherical_joint.py
```

https://user-images.githubusercontent.com/2360366/180027045-d9753a1a-56ec-4459-a719-a13c7398620e.mp4

**Fixed joint:** 

System consisting of two rods and one cylinder connected with fixed joints. The sinusoidal force is applied to the second rod. Example can be run using:

```bash
python examples/JointCases/fixed_joint.py
```

https://user-images.githubusercontent.com/2360366/180449881-c806dd79-324b-40d9-9f57-b6257c7f8978.mp4

## TODOs

- [ ] Write tests for connecting rods with cylinders, particularly for non-zero `point_system_one` and `point_system_two` parameters
- [x] Add cylinder to `spherical_joint.py` example as third element.
- [x] Add cylinder to `fixed_joint.py` example as third element


armantekinalp:
@mstoelzle is the simulation in second video explodes ?

elastica/rod/cosserat_rod.py
```diff
@@ -520,6 +520,164 @@ def compute_shear_energy(self):
...
+            Position of the point in the inertial frame.
+
+        Examples
+        --------
+        Compute position in inertial frame for a point (0, 0, 1) relative to the last node of the rod.
+
+        >>> rod.compute_position_of_point(np.array([0, 0, 1]), -1)
+        """
+        assert point.shape == (3,), "Point must be in the shape (3,)"
+        position = self.position_collection[..., index] + np.dot(
```
armantekinalp, line 550: I still need to look PR, but why do you define the point as an actual point in space instead of some percentage of lengths. Since rod element lengths can change due to extension.

mstoelzle: @armantekinalp Thanks for your question. These points are always in the local coordinate system relative to a node. So when the rod elongates, this point moves with the node. The main idea for usage is for the ends of the rods: the joint might not be attached along the backbone of the rod, but rather touching the cylindrical surface. Or the joint might be displaced by a small distance from the end of a rod through an (unmodelled) rigid element.

bhosale2:
@mstoelzle we can focus on this one after #143 is completed, which is closer to completion.

elastica/rod/cosserat_rod.py
bhosale2: @armantekinalp  thoughts?

bhosale2:
@mstoelzle what's the status here? The PR says WIP do you still need to implement something?

mstoelzle:
> @mstoelzle what's the status here? The PR says WIP do you still need to implement something?

@bhosale2 Although some tests are still missing, I would like to ask you to already have a look at this PR. I would like to prevent writing some tests for functionality / interfaces we might not need / do differently at the end.

bhosale2:
@mstoelzle I went through the PR briefly and noticed that 2 concepts are being implemented simultaneously here:

1. Joint between the rod and rigid body
2. The joint itself is now configurable via the `point_system` feature that you introduce here, where the entities can now be connected at different points, and not constrained to the end of the rigid body or the node of the Cosserat rod element.

In my opinion, the second feature shows some base class structure change, and also should be a separate PR in itself? Because the connection point feature also adds more configurability to joints between rods themselves. Maybe here we stick to the more frequent version of connections at ends?

Let's take a call after seeing what @skim0119 and @armantekinalp think.

mstoelzle:
> @mstoelzle I went through the PR briefly and noticed that 2 concepts are being implemented simultaneously here:
> 
> 
> 
> 1. Joint between the rod and rigid body
> 
> 2. The joint itself is now configurable via the `point_system` feature that you introduce here, where the entities can now be connected at different points, and not constrained to the end of the rigid body or the node of the Cosserat rod element.
> 
> 
> 
> In my opinion, the second feature shows some base class structure change, and also should be a separate PR in itself? Because the connection point feature also adds more configurability to joints between rods themselves. Maybe here we stick to the more frequent version of connections at ends?
> 
> 
> 
> Let's take a call after seeing what @skim0119 and @armantekinalp think.

Thanks @bhosale2 for having a look at the PR. I agree with your analysis that this PR incorporates the two components you mentioned. It is important however to keep in mind the following points:

1. If we want to connect a rod with a cylinder at the nodes (e.g. center of mass of the cylinder - so the traditional way), there is actually only the tiny change in the 'HingeJoint' class necessary. The 'FreeJoint' wouldn't need any modifications.
2. However, while connecting at a node is suitable for Cosserat rods most of the time, I think it is not useful for rigid bodies in a most cases, as we would hardly want the Center of Mass of the Rigid Body to coincide with the node of the Cosserat Rod. The much larger impact can be created when we give the user the design freedom to choose at which point the joint should be attached to the rigid body: for example the bottom or top of a cylinder etc. 
3. Even if we come up with some hard-coded strategy such as that cylinders are always connected at its base at radius zero (e.g. end of cylinder), most of the code changes would still be necessary and the PR would only be slightly smaller.

elastica/rod/cosserat_rod.py
```diff
@@ -520,6 +520,164 @@ def compute_shear_energy(self):
             * (_batch_dot(sigma_diff, shear_internal_forces) * self.rest_lengths).sum()
         )
 
+    def compute_position_of_point(self, point, index):
```
armantekinalp, line 523: Let's remove `compute_position_of_point` and  `compute_velocity_of_point` from Cosserat Rod and include them in joint file as functions.

elastica/rigidbody/rigid_body.py
```diff
@@ -104,3 +104,76 @@ def compute_rotational_energy(self):
             "ijk,jk->ik", self.mass_second_moment_of_inertia, self.omega_collection
         )
         return 0.5 * np.einsum("ik,ik->k", self.omega_collection, J_omega).sum()
+
+    def compute_position_of_point(self, point, index: int = 0):
```
armantekinalp, line 108: Let's remove `compute_position_of_point` and  `compute_velocity_of_point` from RigidBody and include them in joint file as functions.

armantekinalp (changes requested):
1. Lets keep the previous `FreeJoint` implementation as it is and introduce a new class to connect rigid body and Cosserat Rod.
2. I understand why you need to connect cylinder at some other point then the COM, but what is the reason to connect rod a point different than its nodes? Since always discretization of the rod can be changed. 
3. Also please remove `compute_position_of_point` method from Cosserat Rod and Rigid body and include inside the joint.py file as a free function.

mstoelzle:
> * Lets keep the previous `FreeJoint` implementation as it is and introduce a new class to connect rigid body and Cosserat Rod.

As `FixedJoint` is inheriting from `FreeJoint` and also needs to work with rigid bodies, this would require us to do a separate `FixedJoint` class as well to connect rigid bodies. 

- Does this make sense to have this much duplicate code? 
- How do you propose to call the `FreeJoint` and `FixedJoint`, which are able to connect to rigid bodies?

> * I understand why you need to connect cylinder at some other point then the COM, but what is the reason to connect rod a point different than its nodes? Since always discretization of the rod can be changed.

Thanks for your comment. I would like to divide my response into three parts:

- Allowing points as input parameters for both rigid bodies and Cosserat rod nodes makes the implementation easier, the interface cleaner and requires less _if_ statements, as we do not need to distinguish between system types / classes.
- While `point` offsets in z-dir from the Cosserat rod joint can be compensated by choosing a different discretization, as you propose, any radial offset cannot be evoked by choosing a different Cosserat rod configuration, as the mass distribution would be different.
- I currently do not have a specific example / application in mind for defining a point relative to a node. Still, I think it's nice to give the user the freedom to tailor the joint to their use-case, also with regard to connecting Cosserat rods

> * Also please remove `compute_position_of_point` method from Cosserat Rod and Rigid body and include inside the joint.py file as a free function.

Thanks - I will do that.

elastica/rigidbody/rigid_body.py
mstoelzle: Implemented in https://github.com/GazzolaLab/PyElastica/pull/149/commits/74a9c9d3f1511d9aed73a7f7f7b804cf0bebb284

elastica/rod/cosserat_rod.py
mstoelzle: Implemented in https://github.com/GazzolaLab/PyElastica/pull/149/commits/74a9c9d3f1511d9aed73a7f7f7b804cf0bebb284

skim0119 (commented):
@mstoelzle Thank you for your work! I liked your idea and implementation in general :)

I appreciate your effort to make the joint class more generalizable for both rod and rigid body base types. If you want to complete this PR within v0.3, I have a couple of suggestions:

- Lets move `point-system` implementation to the `experimental` folder.
  - Create `experimental/connection_contact_joint/rod_rigidbody_connection.py`
  - We are currently working on refactoring joint class and others (#113), and we don't want to directly merge any code into `joint.py` until it is sorted out.
- If you have any paper or research experiment to validate, please consider sharing it with us. We want to have some confidence in our simulation regarding physical validation.
- In the future with more data, we'll refactor them into the main `joint.py` module.

Other code styles and documentation look good.

@armantekinalp @bhosale2 Thoughts?

armantekinalp:
I think moving this implementation under experiments will help us doing refactoring later on. Similarly, we can have rigid body rod connection in different examples, so future users will have more references to look. So addition to the @skim0119  comments, I would suggest renaming the rigid-body rod connection examples for spherical and fixed joint as, `rod_cylinder_spherical_joint.py` and `rod_cylinder_fixed_joint.py` and have them under different folder.

mstoelzle:
> @mstoelzle Thank you for your work! I liked your idea and implementation in general :)

Thanks everybody for all your help in reviewing and suggesting changes for all the PRs in the last few weeks. I appreciate it and I think it definitely improved the code quality!

> 
> * Lets move `point-system` implementation to the `experimental` folder.
>   
>   * Create `experimental/connection_contact_joint/rod_rigidbody_connection.py`

@skim0119 I have done that in https://github.com/GazzolaLab/PyElastica/pull/149/commits/ff77782ce9b9ffb0c0cf690ce45fdec3d472565b


> So addition to the @skim0119 comments, I would suggest renaming the rigid-body rod connection examples for spherical and fixed joint as, `rod_cylinder_spherical_joint.py` and `rod_cylinder_fixed_joint.py` and have them under different folder.

@armantekinalp I have addressed this in https://github.com/GazzolaLab/PyElastica/pull/149/commits/ff77782ce9b9ffb0c0cf690ce45fdec3d472565b as well

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+from elastica._rotations import _inv_rotate
+from elastica.joint import FreeJoint, FixedJoint
+from elastica.typing import SystemType
+from elastica.utils import Tolerance, MaxDimension
+import numpy as np
+
+
+class RodRigidBodyFreeJoint(FreeJoint):
+    """
+    This free joint class is the base class for all joints. Free or spherical
```
bhosale2, line 14:
```suggestion
    This free joint class is the base class for all joints between a rod and rigid body. Free or spherical
```

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+    k: float
+        Stiffness coefficient of the joint.
+    nu: float
+        Damping coefficient of the joint.
+    point_system_one : numpy.ndarray
+            Describes for system one in the local coordinate system the translation from the node `index_one` (for rods)
+            or the center of mass (for rigid bodies) to the joint.
+    point_system_two : numpy.ndarray
+        Describes for system two in the local coordinate system the translation from the node `index_two` (for rods)
+        or the center of mass (for rigid bodies) to the joint.
```
bhosale2, line 33: shouldn't one of these refer to the rod and the other to the rigid body now? based on the naming?

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+    How to connect the distal end of a CosseratRod with the base of a cylinder using a spherical joint.
+
+    >>> simulator.connect(rod, cylinder, first_connect_idx=-1, second_connect_idx=0).using(
+    ...    FreeJoint,
+    ...    k=1e4,
+    ...    nu=1,
+    ...    point_system_two=np.array([0.0, 0.0, -cylinder.length / 2.]),
+    ... )
+
+    """
```
bhosale2, line 57: modify to reflect joint functionality

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+        nu: float
+           Damping coefficient of the joint.
+        point_system_one : numpy.ndarray
+            Describes for system one in the local coordinate system the translation from the node `index_one` (for rods)
+            or the center of mass (for rigid bodies) to the joint.
+            (default = np.array([0.0, 0.0, 0.0]))
+        point_system_two : numpy.ndarray
+            Describes for system two in the local coordinate system the translation from the node `index_two` (for rods)
+            or the center of mass (for rigid bodies) to the joint.
+            (default = np.array([0.0, 0.0, 0.0]))
```
bhosale2, line 77: ^ditto

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+             System two of the joint connection.
+             Object of a child class of  either `CosseratRod` or `RigidBodyBase`.
+        index_one : int
+            Index of first system for joint.
+        system_two : SystemType
+            System two of the joint connection.
+            Object of a child class of  either `CosseratRod` or `RigidBodyBase`.
+        index_two : int
+            Index of second system for joint.
+
```
bhosale2, line 108: if we know the joint is between a rod and a rigid body maybe like the external contact we can ask for the first body to be a rod and the second to be a rigid body? modify docstring and functionality accordingly?

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+        # the two systems.
+        elastic_force = self.k * end_distance_vector
+
+        # Compute the velocity deviation of the point belonging to system one from the point belonging to system two
+        relative_velocity = velocity_system_two - velocity_system_one
+        # only penalize (e.g. apply opposite damping force) velocity aligned with the `normalized_end_distance_vector`
+        normal_relative_velocity = (
+            np.dot(relative_velocity, normalized_end_distance_vector)
+            * normalized_end_distance_vector
+        )
```
bhosale2, line 158: I don't think we need this projection, we can directly have a resistive force along the direction of relative velocity

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+        # this implementation timed: 2.48 µs ± 126 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
+        end_distance = np.sqrt(np.dot(end_distance_vector, end_distance_vector))
+
+        # Below if check is not efficient find something else
+        # We are checking if the point of `system_one` and point of `system_two` are at the same point in space
+        # If they are at the same point in space, it is a zero vector.
+        if end_distance <= Tolerance.atol():
+            normalized_end_distance_vector = np.array([0.0, 0.0, 0.0])
+        else:
+            normalized_end_distance_vector = end_distance_vector / end_distance
```
bhosale2, line 146: this wont be needed based on the comment below

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+        system_one : SystemType
+             System two of the joint connection.
+             Object of a child class of  either `CosseratRod` or `RigidBodyBase`.
+        index_one : int
+            Index of first system for joint.
+        system_two : SystemType
+            System two of the joint connection.
+            Object of a child class of  either `CosseratRod` or `RigidBodyBase`.
+        index_two : int
+            Index of second system for joint.
```
bhosale2, line 214: modify as mentioned above

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+        kt: float
+            Rotational stiffness coefficient of the joint.
+        nut: float
+            Rotational damping coefficient of the joint.
+        point_system_one : numpy.ndarray
+            Describes for system one in the local coordinate system the translation from the node `index_one` (for rods)
+            or the center of mass (for rigid bodies) to the joint.
+        point_system_two : numpy.ndarray
+            Describes for system two in the local coordinate system the translation from the node `index_two` (for rods)
+            or the center of mass (for rigid bodies) to the joint.
```
bhosale2, line 245: reflect as above

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
```diff
@@ -0,0 +1,417 @@
...
+        nut: float = 0.
+            Rotational damping coefficient of the joint.
+        point_system_one : numpy.ndarray
+            Describes for system one in the local coordinate system the translation from the node `index_one` (for rods)
+            or the center of mass (for rigid bodies) to the joint.
+            (default = np.array([0.0, 0.0, 0.0]))
+        point_system_two : numpy.ndarray
+            Describes for system two in the local coordinate system the translation from the node `index_two` (for rods)
+            or the center of mass (for rigid bodies) to the joint.
+            (default = np.array([0.0, 0.0, 0.0]))
```
bhosale2, line 306: ^ditto

bhosale2 (changes requested):
Minor comments, nice work overall!

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
mstoelzle: @bhosale2 I have a different opinion than you on this one. I want to leave the option open for the user to combine systems in any way they would like:

- **Rod to Rod:** In this situation, the users can also use the standard (non-experimental) `FreeJoint` implementation, which should numerically give the same result, but is possibly computationally slightly faster
- **Rod to Rigid Body**
- **Rigid Body to Rod**: This should have the same exact effect as the second bullet point, provided that the user mirrors the configuration.
- **Rigid Body to Rigid Body**: Most importantly, this class should also enable this configuration. If I force the first system to be a rod and the second system to be a rigid body, this is not possible anymore and not implemented anywhere else

I am open though to changing the name of the file / of the classes if you think something would be more intuitive than the current naming scheme.

bhosale2: I appreciate the generality seen in the code here, but it is not consistent with the naming of the function then, which when seen one expects a joint between a rod and rigid body. In that case something like a generic_joint would make more sense? generalised_joint? @armantekinalp can voice his opinion as well on this

mstoelzle: @bhosale2 I see those two classes as a more general implementation of the original FreeJoint and FixedJoint, which should be called something like Rod2RodFreeJoint and Rod2RodFixedJoint in the future. 

So I would be actually in favor in calling the new, more general implementation either FreeJoint and FixedJoint or NewFreeJoint and  NewFixedJoint to communicate to the user that this will be (most likely) the future standard implementation of these joint classes.

bhosale2: I agree with the first part, but refactoring the core codebase `FreeJoint` has its own drawbacks. `NewFreejoint` may be a good idea. Anyways, for this PR we are shifting these to experimental as planned, and we can revisit this naming then later in the future.

mstoelzle: @armantekinalp If you agree as well, I will rename it accordingly to NewFreeJoint and NewFixedJoint.  They are in the experimental module anyway.

armantekinalp: Hi @mstoelzle . After discussing with @bhosale2 and @skim0119,  we think we can rename them as `GenericSystemTypeFixedJoint` and `GenericSystemTypeFreeJoint`for this PR.

armantekinalp (changes requested):
Minor changes.

tests/test_joint.py
```diff
@@ -207,12 +221,9 @@ def test_hingejoint():
         rod2.external_forces[..., rod2_index], -1 * contactforce, atol=Tolerance.atol()
     )
 
-    linkdirection = (
-        rod2.position_collection[..., rod2_index + 1]
-        - rod2.position_collection[..., rod2_index]
-    )
```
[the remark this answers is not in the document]
mstoelzle: @armantekinalp I changed the computation of the link direction so that it is compatible with rigid bodies. The computation using the position of multiple nodes only works for rods.

examples/JointCases/fixed_joint.py
```diff
@@ -75,7 +81,13 @@ class FixedJointSimulator(
 # Connect rod 1 and rod 2
 fixed_joint_sim.connect(
     first_rod=rod1, second_rod=rod2, first_connect_idx=-1, second_connect_idx=0
-).using(FixedJoint, k=1e5, nu=0, kt=5e3)
+).using(
+    FixedJoint,
+    k=1e5,
+    nu=0e0,
+    kt=1e1,
```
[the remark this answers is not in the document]
mstoelzle: @armantekinalp I actually badly tuned this spring constant for the torsion before in the previous PR. The example simulation was very unstable and fragile as it would disconverge often. 

As this spring is now directly acting on the rotation instead some position difference (as changed in previous PR), a lower spring constant seems to be suitable.

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
bhosale2: @mstoelzle  updates on these?

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
bhosale2: updates here @mstoelzle ?

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
mstoelzle: @bhosale2 Sorry, I missed this comment. GitHub issues are sometimes a bit messy...

I think I am the wrong person to ask about this. I literally copied this from the original 'FreeJoint' implementation. 

Out of my intuition, I don't think this projection is necessary. Please let me know if I can remove it.

bhosale2: I just confirmed with @armantekinalp, you can go ahead.

elastica/experimental/connection_contact_joint/rod_rigidbody_connection.py
mstoelzle: Thanks for this suggestion. I renamed in https://github.com/GazzolaLab/PyElastica/pull/149/commits/002770e757697ff071a419358aa75af4d0b9839c

examples/JointCases/fixed_joint.py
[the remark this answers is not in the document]
armantekinalp: okay thanks for clarifying.

armantekinalp (approved):
LGTM

docs/api/connections.rst
```diff
@@ -26,8 +26,10 @@ Compatibility
 Connection / Contact / Joints   Rod  Rigid Body
 =============================== ==== =========== 
 FreeJoint                       ✅   ❌
+GenericSystemTypeFreeJoint      ✅   ✅
 ExternalContact                 ✅   ❌
 FixedJoint                      ✅   ❌
+GenericSystemTypeFixedJoint     ✅   ✅
```
bhosale2, line 32: please remove from docs, we don't host experimental features here yet.

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,414 @@
...
+        end_distance = np.sqrt(np.dot(end_distance_vector, end_distance_vector))
+
+        # Below if check is not efficient find something else
+        # We are checking if the point of `system_one` and point of `system_two` are at the same point in space
+        # If they are at the same point in space, it is a zero vector.
+        if end_distance <= Tolerance.atol():
+            normalized_end_distance_vector = np.array([0.0, 0.0, 0.0])
+        else:
+            normalized_end_distance_vector = end_distance_vector / end_distance
+
```
bhosale2, line 146:
```suggestion
```

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,414 @@
...
+        elastic_force = self.k * end_distance_vector
+
+        # Compute the velocity deviation of the point belonging to system one from the point belonging to system two
+        relative_velocity = velocity_system_two - velocity_system_one
+        # only penalize (e.g. apply opposite damping force) velocity aligned with the `normalized_end_distance_vector`
+        normal_relative_velocity = (
+            np.dot(relative_velocity, normalized_end_distance_vector)
+            * normalized_end_distance_vector
+        )
+        damping_force = -self.nu * normal_relative_velocity
```
bhosale2, line 158:
```suggestion
        damping_force = -self.nu * relative_velocity
```

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,414 @@
...
+        1D (3,) numpy array containing 'float' data.
+        Position of the point in the inertial frame.
+
+    Examples
+    --------
+    Compute position in inertial frame for a point (0, 0, 1) relative to the last node of the rod.
+
+    >>> system.compute_position_of_point(np.array([0, 0, 1]), -1)
+    """
+    assert point.shape == (3,), "Point must be in the shape (3,)"
```
bhosale2, line 363: Maybe not a good idea to have an `assert` in a function that will be called every time step?

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,414 @@
...
+        1D (3,) numpy array containing 'float' data.
+        Velocity of the point in the inertial frame.
+
+    Examples
+    --------
+    Compute velocity in inertial frame for a point (0, 0, 1) relative to the last node of the rod.
+
+    >>> system.compute_velocity_of_point(np.array([0, 0, 1]), -1)
+    """
+    assert point.shape == (3,), "Point must be in the shape (3,)"
```
bhosale2, line 399: ^ditto

bhosale2 (changes requested):
Minor stuff, good work @mstoelzle

docs/api/connections.rst
mstoelzle: Ok, I see. I removed it in https://github.com/GazzolaLab/PyElastica/pull/149/commits/63be1a41d9aac93826a52d3171766d4984ab2d45

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
mstoelzle: Done in https://github.com/GazzolaLab/PyElastica/pull/149/commits/a01136f57363afdc4b835ad163283cbd8926b343

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
mstoelzle: Removed in https://github.com/GazzolaLab/PyElastica/pull/149/commits/a01136f57363afdc4b835ad163283cbd8926b343

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
mstoelzle: Implemented in https://github.com/GazzolaLab/PyElastica/pull/149/commits/cc2010ddebfff9cf7128930dfb6e6e9ad92eb167

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
mstoelzle: Implemented in https://github.com/GazzolaLab/PyElastica/pull/149/commits/cc2010ddebfff9cf7128930dfb6e6e9ad92eb167

mstoelzle:
> 



> Minor stuff, good work @mstoelzle, after this round it will be LGTM from my side.
> Also based on #163, please pull latest `update`, then modify the rod damper class name `ExponentialDamper` to `AnalyticalLinearDamper`.

Thanks for your suggestions. I addressed all of them and I also fixed a bug in the changes for the `HingeJoint`

examples/JointCases/experimental/generic_system_type_fixed_joint.py
```diff
@@ -0,0 +1,205 @@
+__doc__ = """Fixed joint example, for detailed explanation refer to Zhang et. al. Nature Comm.  methods section."""
+
+import numpy as np
+import sys
+
+# FIXME without appending sys.path make it more generic
+sys.path.append("../../")
```
bhosale2, line 7:
```suggestion
```

examples/JointCases/experimental/generic_system_type_spherical_joint.py
```diff
@@ -0,0 +1,201 @@
+__doc__ = """Spherical(Free) joint example, for detailed explanation refer to Zhang et. al. Nature Comm.  
+methods section."""
+
+import numpy as np
+import sys
+
+# FIXME without appending sys.path make it more generic
+sys.path.append("../../")
```
bhosale2, line 8:
```suggestion
```

bhosale2 (changes requested):
Super minor stuff.

bhosale2 (approved):
LGTM. @skim0119 please take over.

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,396 @@
...
+
+    def apply_forces(self, system_one, index_one, system_two, index_two):
+        """
+        Apply joint force to the connected systems.
+
+        Parameters
+        ----------
+        system_one : SystemType
+             System two of the joint connection.
+             Object of a child class of  either `CosseratRod` or `RigidBodyBase`.
```
skim0119, line 99:
```suggestion
```

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,396 @@
...
+        index_one : int
+            Index of first system for joint.
+        system_two : SystemType
+            System two of the joint connection.
+            Object of a child class of  either `CosseratRod` or `RigidBodyBase`.
```
skim0119, line 104:
```suggestion
```

skim0119 (changes requested):
@mstoelzle LGTM. Minor changes for documentation consistency.

@armantekinalp Do you think we should make a `nightly` doc page that includes experimental modules?

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,396 @@
...
+           Damping coefficient of the joint.
+        point_system_one : numpy.ndarray
+            Describes for system one in the local coordinate system the translation from the node `index_one` (for rods)
+            or the center of mass (for rigid bodies) to the joint.
+            (default = np.array([0.0, 0.0, 0.0]))
+        point_system_two : numpy.ndarray
+            Describes for system two in the local coordinate system the translation from the node `index_two` (for rods)
+            or the center of mass (for rigid bodies) to the joint.
+            (default = np.array([0.0, 0.0, 0.0]))
+        """
```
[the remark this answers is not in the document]
mstoelzle: Addressed in https://github.com/GazzolaLab/PyElastica/pull/149/commits/f5f4d683adb72aff7c13bd30f4a07398f75ba74b

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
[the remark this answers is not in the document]
mstoelzle: Sorry, I am not quite sure what you mean here. Could you please go into more detail? Possibly it is already resolved by the latest changes

tests/test_generic_system_type_joint.py
```diff
@@ -0,0 +1,277 @@
+__doc__ = """ Joints for generic system types test module """
```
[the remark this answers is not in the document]
mstoelzle: Done in https://github.com/GazzolaLab/PyElastica/pull/149/commits/d64072e11f24a365febbd26c474fa1fdd6afc0c3

tests/test_generic_system_type_joint.py
```diff
@@ -0,0 +1,277 @@
...
+from numpy.testing import assert_allclose
+from elastica.rigidbody import Cylinder, Sphere, RigidBodyBase
+from elastica.rod.cosserat_rod import CosseratRod
+from elastica.utils import Tolerance
+import numpy as np
+import pytest
+from typing import *
+
+# seed random number generator
+np.random.seed(0)
```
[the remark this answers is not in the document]
mstoelzle: Done in https://github.com/GazzolaLab/PyElastica/pull/149/commits/ef5d5c1ae7e0598a791e650c584d95554a9c42a7

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,396 @@
...
+        )
+
+    # Apply force is same as free joint
+    def apply_forces(self, system_one, index_one, system_two, index_two):
+        return GenericSystemTypeFreeJoint.apply_forces(
+            self, system_one, index_one, system_two, index_two
+        )
+
+    def apply_torques(self, system_one, index_one, system_two, index_two):
+        return FixedJoint.apply_torques(
```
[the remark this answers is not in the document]
mstoelzle: `super()` wouldn't work here, as I need to access the `apply_torques` method of the `FixedJoint` class and I need to specify that I do not want to access the `apply_torques` method of the `GenericSystemTypeFreeJoint` class instead

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,396 @@
...
+            Index of second system for joint.
+
+        Returns
+        -------
+
+        """
+        pass
+
+
+class GenericSystemTypeFixedJoint(GenericSystemTypeFreeJoint, FixedJoint):
```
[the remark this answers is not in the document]
mstoelzle: @skim0119 Thanks for your comment. I don't think your suggestion would work, as we need to access the `apply_torques` method of the `FixedJoint` class. We also need to inherit from `GenericSystemTypeFreeJoint` to access their `apply_forces` method. This `GenericSystemTypeFixedJoint` class is basically a combination of `GenericSystemTypeFreeJoint` and `FixedJoint`.

Please let me know if you have another idea how to solve this.

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,396 @@
...
+
+        # Compute elastic force using a spring formulation as a linear function of the (undesired) distance between
+        # the two systems.
+        elastic_force = self.k * distance_vector
+
+        # Compute the velocity deviation of the point belonging to system one from the point belonging to system two
+        relative_velocity = velocity_system_two - velocity_system_one
+
+        # Compute damping force considering the specified damping coefficient `nu`
+        damping_force = self.nu * relative_velocity
```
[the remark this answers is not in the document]
mstoelzle: @skim0119 I actually looked into this topic yesterday again, but I would love to verify my thoughts with you guys:

We can formulate the problem such that we look at system one and consider the state of system two the desired (kinematic) goal state for system one. Based on the common Robotics literature such as  the book by Siciliano, we formulate our kinematic PD regulator as:

f = k_p (x_des - x) + k_d (v_des - v)

Where k_p and k_d are the proportional and damping gains respectively, x_des the desired position and v_des the desired velocity. 

Now we can adapt this formulation to our problem, as explained above:

f_1 = k (x_2 - x_1) + nu (v_2 - v_1)

This force f_1 will guide system one towards system two. 
Then, we apply the opposite force to system two to guide it towards the state of system one.

What I found though when running the spherical joint example is that this damping constant nu cannot be chosen too big, as the regulator becomes unstable. I assume that this comes from the rapidly changing "goal" state.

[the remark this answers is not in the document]
bhosale2: @mstoelzle thanks for explaining this, we actually noticed some inconsistencies on our end (described in #164). Although can you please set some small value of nu for the joints and confirm if it runs as expected? Maybe with a stabilized behavior? You don't need to set a high value.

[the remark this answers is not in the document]
mstoelzle: @bhosale2 I haven't necessary verified how the damping verifies the multi-body system characteristics (e.g. whether it actually removes energy from the system), but I tested that the `generic_system_type_spherical_joint` example works without issues with a `k=1e3` and `nu=1e-1`. I didn't invest too much time on tuning these two values, but at some point when you increase either `k` or `nu` too much, the system becomes unstable.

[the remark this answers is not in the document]
bhosale2: Ok that behavior is expected thanks @mstoelzle

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,421 @@
...
+        ... )
+    """
+
+    def __init__(
+        self,
+        k: float,
+        nu: float,
+        kt: float,
+        nut: float = 0.0,
+        point_system_one: np.ndarray = None,
```
skim0119, line 279: Add `Optional`. Also, please don't put the default value here, the default value should be described below (as you did 👍 . It's for documentation esthetic, so please understand.)

Optional already implies the default is None.

skim0119 (changes requested):

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
```diff
@@ -0,0 +1,421 @@
...
+    """
+
+    # pass the k and nu for the forces
+    # also the necessary systems for the joint
+    def __init__(
+        self,
+        k: float,
+        nu: float,
+        point_system_one: np.ndarray = None,
+        point_system_two: np.ndarray = None,
```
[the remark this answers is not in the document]
mstoelzle: Python does not automatically initialize the parameter to the default value as hinted by the typing specification, see https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint

That is why we still need to set the default value to None, similar to https://github.com/GazzolaLab/PyElastica/blob/master/elastica/rod/factory_function.py#L23

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
mstoelzle: Python does not automatically initialize the parameter to the default value as hinted by the typing specification, see https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint

That is why we still need to set the default value to None, similar to https://github.com/GazzolaLab/PyElastica/blob/master/elastica/rod/factory_function.py#L23

mstoelzle: Also, starting at Python 3.9 the `Optional` typing hint is not necessary anymore, see https://stackoverflow.com/a/67370289/1745305

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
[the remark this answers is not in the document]
mstoelzle: Fixed in https://github.com/GazzolaLab/PyElastica/pull/149/commits/f907e1ad86eb30b9a3bb2c8affee4d15ac2fc149

elastica/experimental/connection_contact_joint/generic_system_type_connection.py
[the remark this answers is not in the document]
mstoelzle: > I understand your point, but I feel it's little dangerous using multiple inheritances here. Those modules are not designed to be used as a mixin, and it can cause issues downstream.
> 
> For now, if you just want to assign the same method, simply put
> 
> ```python
> apply_torques = FixedJoint.apply_toruqes
> apply_forces = GenericSystemTypeFreeJoint.apply_forces
> ```
> 
> inside the class, instead of inheriting. You can also call `__init__` without inheriting, so your `__init__` method doesn't have to be changed.
> 
> > As you can see, the inheritance doesn't mean anything in your current implementation: it has a chance to be misinterpreted in later usage. Directly calling `GenericSystemTypeFreeJoint` and `FixedJoint` doesn't mean you are calling the parent entities; they are just calling other class members as static.
> 
> It appears to me that this class will eventually be converted to `FixedJoint` in the future, so let's keep the code as intuitive as possible.

This doesn't work, as the `__init__` cannot be called without inheriting. For the code
```py
FixedJoint.__init__(
            self, k=k, nu=nu, kt=kt, nut=nut, rest_rotation_matrix=rest_rotation_matrix
        )
```
I get the following error:
```
Traceback (most recent call last):
  File "/Users/maximilianstoelzle/Documents/phd/PyElastica/elastica/wrappers/connections.py", line 278, in __call__
    return self._connect_cls(*self._args, **self._kwargs)
  File "/Users/maximilianstoelzle/Documents/phd/PyElastica/elastica/experimental/connection_contact_joint/generic_system_type_connection.py", line 318, in __init__
    FixedJoint.__init__(
  File "/Users/maximilianstoelzle/Documents/phd/PyElastica/elastica/joint.py", line 240, in __init__
    super().__init__(k, nu)
TypeError: super(type, obj): obj must be an instance or subtype of type

```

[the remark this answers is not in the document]
mstoelzle: @skim0119  In my mind we have four options:

1. You know of a solution how to call the (__init__) methods if non-inherited classes 
2. We leave the inheritance like it is now. I think the syntax is quite clear how we leverage the functionality of both base classes we inherit from. I don't think it will cause any issues as a) it is an experimental feature anyway, so interface and functionality changes need to be expected b) As you mentioned the code should hopefully be merged / replace the FixedJoint class. 
3. We just copy the necessary code from the FixedJoint class into the GenericSystemTypeFixedJoint class. 
4. We created a static method in the FixedJoint class, which can also be called from the GenericSystemTypeJoint class without inheriting.

@skim0119 Please let me know which way you want to go.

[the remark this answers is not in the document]
skim0119: @mstoelzle 
I guess its a bit tricky. For now, let's just copy the code for `__init__`. I'll clear the code when we merge to `joint.py`. In the end, I think the sequence needs to be modularized.

For `apply_force` and `apply_torques`, I think it shouldn't cause an issue, right?

## Suggestions:

- Use `FreeJoint` as a base
- Have its own `__init__` function
- Make alias for `apply_force` and `apply_torques`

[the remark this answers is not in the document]
mstoelzle: > @mstoelzle I guess its a bit tricky. For now, let's just copy the code for `__init__`. I'll clear the code when we merge to `joint.py`. In the end, I think the initial computation needs to be modularized.
> 
> For `apply_force` and `apply_torques`, I think it shouldn't cause an issue, right?
> 
> ## Suggestions:
> * Use `FreeJoint` as a base
> * Have its own `__init__` function
> * Make alias for `apply_force` and `apply_torques`

@skim0119 I removed the inheritance from `FixedJoint` in https://github.com/GazzolaLab/PyElastica/pull/149/commits/80d0ed1511d41f457b4f6eb5994f6d0a3f8b661e. The `GenericSystemTypeFixedJoint` now uses `GenericSystemTypeFreeJoint` as a base class so that we can use their `__init__` without copying the code. The necessary `__init__` functionality from `FixedJoint` is copied.

skim0119 (approved):
Looks good. I haven't run the test and examples, but we'll check them before releasing v0.3

PR #160, opened 2022-07-28
update: matplotlib and ffmpeg dependencies

bhosale2:
- makes `matplotlib` an optional dependency required for examples
- remove `ffmpeg` install from `pip` (which doesn't work) and adds appropriate `conda install` command


armantekinalp (approved):
LGTM

PR #162, opened 2022-07-30
update: remove sys append calls in examples

bhosale2:


examples/BoundaryConditionsCases/general_constraint_allow_yaw.py
```diff
@@ -37,7 +32,7 @@ class GeneralConstraintSimulator(
 shear_modulus = E / (poisson_ratio + 1.0)
 
 # setting up timestepper and video
-final_time = 10
+final_time = 1
```
armantekinalp, line 35: why did we changed final time here?

bhosale2: wanted to check if it runs as expected quickly, no other reason. reverted back in 3e7fb50

armantekinalp (approved):
LGTM

PR #172, opened 2022-08-07
update: Damping values for rod-rigid body contact cases (#171)

### LINKED ISSUE #171: Update internal damping values for rod-rigid body contact cases
Since rod-rigid body contact cases were developed in `v0.2.4`, the internal damping values are based on the old model. If possible before the next release, the values need to be updated to be consistent with the new model, to simulate expected dynamics.

bhosale2:
Fixes #171.


armantekinalp (approved):
LGTM

PR #175, opened 2022-08-16
(#165) feat: autoflake funtionality

### LINKED ISSUE #165: Autoflake as a developer dependency for cleaning up unused imports and variables
PyElastica is muddled at a couple of places with unused imports and variables. [`autoflake`](https://github.com/PyCQA/autoflake) can be a good solution as a dev dependency, cleaning up the unused imports and maintaining it that way for future development. For an example implementation in a related project see [MagnetoPyElastica](https://github.com/armantekinalp/MagnetoPyElastica)

bhosale2:
Fixes #165.


skim0119:
I'm honestly didn't have good past experience with `autoflake`. Many intermediate code-push are experimental and work-in-progress, maybe a draft or under review/revision. I preferred `flake8` and `black` because they only touch the formatting without altering the contents itself, but `autoflake` actually changed the contents. It was a bit more than just a `formatting` for me.

I'm not sure, and maybe they made it better nowadays. @bhosale2 Any thought? Is there a way to run the check only before merging?

bhosale2:
@skim0119 I totally understand your concern. That's why I haven't run the formatting yet. Also about the check, the command `make check-codestyle` will actually show you changes that `autoflake` will do before formatting (like `git diff`), and then runs a command that fails if some changes were required by `autoflake`.

Screen Shot 2022-08-16 at 9 45 12 AM

On the other hand, the `make autoflake` command is the one that will actually modify files in place.

If you want to ensure safety w.r.t to code changing, should we then only have `autoflake` in format check and not the actual formatter?  Then the user needs to take care of unused imports/vars to fix the check.

skim0119:
@bhosale2 Hmm.. how about we only include it in CI? Its developer's responsibility to manage their code while they are developing, but anything that goes to shared branch needs to pass formatted. Although, it reminds me when we didn't have pre-commit for black, and had bunch of commit message saying "black formatting" lol

bhosale2:
Exactly that's what I meant, we only add it in `make check-codestyle` which is the CI check, and let the user do the fixing in a sane fashion.

Screen Shot 2022-08-16 at 10 08 41 AM

Ok I will remove `autoflake` from the `make formatting` command then., and fix the formatting changes myself.

armantekinalp:
CI idea sounds good to me, as @skim0119 said developer has to format their code.

skim0119:
Can we also include `# pragma: no autoflake` feature? 👍

bhosale2:
An update here (after discussion with @skim0119), `autoflake` seems to remove `pass` statements after docstrings in empty functions, which are usually important to convey intent. There is some recent ongoing work to allow configurability here, so we will keep this as WIP and fix it in the future when that feature comes in.

skim0119:
^

PRs to look for
- https://github.com/fsouza/autoflake8/pull/70

skim0119 (approved):
LGTM

PR #205, opened 2022-10-15
remove python 3.7 from CI

armantekinalp:
Removing python 3.7 from CI since it creates some conflicts with some packages.


armantekinalp:
I am closing this PR. Lets finish PR #206

PR #206, opened 2022-10-19
Bug fix main yml windows python version

armantekinalp:
- Removes Windows build from CI, due to the bug in CI system.
- Removes Python version 3.7 builds, since some of the dependencies are not supporting 3.7 anymore.

We could activate windows builds in future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,15 +20,15 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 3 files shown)


skim0119:
> * Removes Windows build from CI, due to the bug in CI system.
> * Removes Python version 3.7 builds, since some of the dependencies are not supporting 3.7 anymore.

Which dependencies? If we are dropping support for 3.7, we need to edit `pyproject` as well.

> We could activate windows builds in future.

Can we make an issue for this?

armantekinalp:
New Numba version 0.56.3 is failing. I updated `pyproject` and `poetry`. Also I created a new issue for windows CI.

bhosale2 (approved):
LGTM

PR #232, opened 2023-02-24
# 169: Replaced 'moviepy' with 'ffmpeg' for video generation

Rahul-JOON:
Replaced the video generation functionality from moviepy to ffmpeg because it is much faster and makes it uniform among other examples.

Ps: The codebase was failing some callbacks checks on my local machine, but I believe they are unrelated to my modification.


armantekinalp (approved):
LGTM! Thanks

bhosale2 (approved):
LGTM

PR #247, opened 2023-05-06
changed FixedConstraint example

erfanhamdi:
hey there

The example on the FixedConstraint didn't work for me and as I read the source it seemed that it needed a tuple as the node idx position instead of an int, so I just added a ',' at the end of it and it got fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <div align='center'>
 <h1> PyElastica </h1>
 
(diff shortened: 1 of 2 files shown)


skim0119:
@erfanhamdi Hi -- Thank you for bringing this to our attention. I will merge it as part of documentation patches for now, and merge it into the main on the next update.

PR #249, opened 2023-05-11
239 fix restart functionality

### LINKED ISSUE #239: Possible Bug: Test restart functionality and applying constrains
We should test the below line in the restart code. We suspect it creates incorrect results, when we load and time-step.

https://github.com/GazzolaLab/PyElastica/blob/7b3fe2d0204a175c9e5e64265c70b9577b4ca8c9/elastica/restart.py#L92-L93

armantekinalp:
- Fixes #239 
- Improving  unit tests for restart functions.

We ran two different simulations to test if restart working, dynamic cantilever and ring rod under gravity. Nodal acceleration in time are compared with restart and without restart. 

Dynamic cantilever beam

Dynamic_cantilever_visualization_accelerations

Ring rod under gravity

ring_rod_middle_node_acceleration_testing_restart

### DIFF
--- a/elastica/__init__.py
+++ b/elastica/__init__.py
@@ -71,3 +71,4 @@ from elastica.timestepper import (
 )
 from elastica.memory_block.memory_block_rigid_body import MemoryBlockRigidBody
 from elastica.memory_block.memory_block_rod import MemoryBlockCosseratRod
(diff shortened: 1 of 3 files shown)


bhosale2 (approved):
LGTM.

Issue #99, opened 2022-06-01
Unit tests for contact module


armantekinalp:
**Is your feature request related to a problem? Please describe.**
Although we have tested contact module through different simulations, there is no unit tests for contact module. This can cause vulnerability in development process as well as reduces code coverage. 

**Describe the solution you'd like**
Implement unit tests for contact module.

armantekinalp:
Here are the list of functions that needs to be tests

- Common functions (1 PR)
    - [ ] ` _dot_product`
    - [ ] `_norm`
    - [ ] `_clip`
    - [ ] `_out_of_bounds`
    - [ ] `_find_min_dist`
    - [ ] `_aabbs_not_intersecting`

- More specific functions (2-3 PR)
    - [ ] `_prune_using_aabbs_rod_rigid_body`
    - [ ] `_prune_using_aabbs_rod_rod`
    - [ ] `_calculate_contact_forces_rod_rigid_body`
    - [ ] `_calculate_contact_forces_rod_rod`
    - [ ] `_calculate_contact_forces_self_rod`
    - [ ] 
- Wrapper classes (2 PR)
    - [ ] `ExternalContact`
    - [ ] `SelfContact`

Rahul-JOON:
The specific functions are now complete, I will do the wrapper classes next.

armantekinalp:
This issue is resolved, I am closing.

PR #260, opened 2023-05-30
Unit test for specific functions part-2

Rahul-JOON:
This PR contains the test case for the function - _calculate_contact_forces_rod_rigid_body as mentioned in #99.

Because the codebase is not broken, all of the test cases are currently passing.

### DIFF
--- a/tests/test_contact_specific_functions.py
+++ b/tests/test_contact_specific_functions.py
@@ -3,7 +3,12 @@ __doc__ = (
 )
 
 import numpy as np
-from elastica.joint import _prune_using_aabbs_rod_rigid_body, _prune_using_aabbs_rod_rod
+from numpy.testing import assert_allclose
+from elastica.joint import (
+    _prune_using_aabbs_rod_rigid_body,
+    _prune_using_aabbs_rod_rod,
+    _calculate_contact_forces_rod_rigid_body,
+)
 
 
 class TestPruneUsingAABBSRodRigidBody:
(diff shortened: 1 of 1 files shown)


armantekinalp (approved):
LGTM

armantekinalp (approved):
LGTM thanks

tests/test_contact_specific_functions.py
```diff
@@ -106,3 +111,126 @@ def test_prune_using_aabbs_rod_rod(self):
             )
             == 1
         )
+
+
+class TestCalculateContactForcesRodRigidBody:
+    "class to test the calculate contact forces rod rigid body function"
+
+    def test_claculate_contact_forces_rod_rigid_body(self):
+        "Testing function with analytically verified values"
```
bhosale2, line 120: I'm a bit confused why there are functions inside classes, which do not use the object-oriented nature of the classes. Can these be free functions? Let me know if I am missing anything

bhosale2: Similar comment for other similar places.

bhosale2 (changes requested):

tests/test_contact_specific_functions.py
Rahul-JOON: You are absolutely right; "Pytests may be written either as functions or as methods in classes – unlike unittest, which forces tests to be inside classes." (source: [Python testing](https://automationpanda.com/2017/03/14/python-testing-101-pytest/#:~:text=Pytests%20may%20be%20written%20either,tests%20to%20be%20inside%20classes.) )

The only logical use case of pytest in classes are:

- Joint parametrization of multiple test methods belonging to a given class.
- Reuse of test data and test logic via subclass inheritance.

(discussion: [Stack overflow](https://stackoverflow.com/questions/50016862/grouping-tests-in-pytest-classes-vs-plain-functions) )

But since we are not using common inputs between different functions; the classes provide no more value than just grouping the test cases for a function together. (kinda like division decorator in HTML)

Should I Remove all the classes and push again? @bhosale2

bhosale2: thanks, for the details @Rahul-JOON. Yes please do that.

Rahul-JOON: I implemented the changes in both test_contact_common_fucntions.py as well as test_contact_specific_functions.py; please check.

Issue #270, opened 2023-06-19
Mesh Object Class


Ali-7800:
### Initialization
- [ ] Can be initialized by reading a STL or OBJ mesh file formats.

### Class Attributes
- [ ] faces, dimension: (3 spatial coordinates, 3 vertices, n faces)
This should store the coordinates of the 3 vertices of each of the n faces of the imported mesh.

- [ ] face normals, dimension: (3 spatial coordinates, n faces)
This should store the coordinates of the unit normal vector of each of the n faces.

- [ ] face centers,  dimension: (3 spatial coordinates, n faces)
This should store the coordinates of the position vector of each of the n face centers (average of the vertices).

- [ ] mesh orientation, dimension: (3 spatial coordinates, 3 vectors)
This should store the 3 orthonormal basis vectors that define the mesh orientation, by default this should be the cartesian orthonormal basis vectors.

- [ ] mesh scale, dimension (3 lengths,)
This should store the 3 dimensions of the smallest box that could fit the mesh whose sides align with the mesh orientation vectors.

- [ ] mesh center,  dimension: (3 spatial coordinates,)
This should store the coordinates of the position vector of the center of the smallest box that could fit the mesh.

### Class Methods
- [ ] Method for visualizing the mesh object
- [ ] Method for translating the mesh
Given a position vector; translate the faces, face centers, and mesh center.
- [ ] Method for reorienting the mesh
Given a rotation axis (unit vector) and an angle; rotate the faces, face centers, face normals, and mesh center.
- [ ] Method for rescaling the mesh
Given a scale value; scale the faces, face centers, and mesh scale.

Ali-7800:
@Rahul-JOON can you work on this issue

Rahul-JOON:
Sure, I will take this on.

PR #275, opened 2023-06-29
Mesh init 5/6 attributes

Rahul-JOON:
This PR sets up the base for the mesh class in pyelastica as per issue #270 

This PR contains basic code for reading an stl file and testing for the faces array's geometry that is generated.


mesh/mesh_initializer.py
```diff
@@ -0,0 +1,40 @@
...
+    How to initialize a mesh?
+    -------------------------
+
+    mesh = MeshInitialize(r('<filepath>'))
+
+    PS: Please be sure to add .stl / .obj at the end of the filepath, if already present, ignore.
+    """
+    def __init__(self, filepath):
+        self.mesh = pv.read(filepath)
+        self.pyvista_faces = self.mesh.faces
```
armantekinalp, line 21: why do you name as pyvista_faces and not faces?

mesh/mesh_initializer.py
```diff
@@ -0,0 +1,40 @@
...
+        self.number_of_faces = self.mesh.n_faces
+        self.pyvista_points = self.mesh.points
```
armantekinalp, line 23: Same as above

assets/cube.stl
armantekinalp, line 1: I will get back to you where to keep stl file

armantekinalp (changes requested):
Some minor questions/comments.

mesh/mesh_initializer.py
Rahul-JOON: I named it pyvista_faces/points to specify that it's the faces array from pyvista library, I later used this array and established a self.faces array that stores the data in the desired format and dimension as requested by Ali in #270.

assets/cube.stl
Rahul-JOON: Okay, Thanks!

mesh/mesh_initializer.py
Rahul-JOON: I have added a detailed explanation of the function and class works as comments in the next commit; I will push that soon after this PR is resolved/merged.

assets/cube.stl
armantekinalp: @Rahul-JOON please move this file inside the tests.

armantekinalp: Also is stl files always binary? Shouldn't it be readable

Rahul-JOON: Sure, I will move the file.
No, the stl file comes in both a human-readable ASCII format or a smaller binary file format; to reduce the file size, I chose the binary one.

Ali-7800 (approved):
LGTM

mesh/mesh_initializer.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    This Mesh Initializer class uses pyvista to import mesh files in the
+    STL or OBJ file formats and initializes the necessary mesh information.
+
+    How to initialize a mesh?
+    -------------------------
+
+    mesh = MeshInitialize(r('<filepath>'))
+
+    PS: Please be sure to add .stl / .obj at the end of the filepath, if already present, ignore.
```
bhosale2, line 17: Add this under Notes and not PS

mesh/mesh_initializer.py
```diff
@@ -0,0 +1,40 @@
+__doc__ = """ Mesh Initializer using Pyvista """
+
+import pyvista as pv
+import numpy as np
+
+
+class MeshInitialize():
```
bhosale2, line 7: why call it `MeshInitialize` and not just `Mesh`?

mesh/mesh_initializer.py
```diff
@@ -0,0 +1,40 @@
...
+        self.mesh = pv.read(filepath)
+        self.pyvista_faces = self.mesh.faces
+        self.number_of_faces = self.mesh.n_faces
+        self.pyvista_points = self.mesh.points
+        self.faces = self.face_calculation(self.pyvista_faces, self.pyvista_points, self.number_of_faces)
+
+    def face_calculation(self, pvfaces, meshpoints, n_faces):
+        """
+        This function converts the faces from pyvista to pyelastica geometry
+        """
```
bhosale2, line 29: @armantekinalp  I was thinking that here on wards we should try to make it a habit of adding type hints and including the information in doc string as needed. Ofc eventually we need to propagate this backward in `pyelastica` as well. Thoughts?

bhosale2 (changes requested):
Minor comments

mesh/mesh_initializer.py
Rahul-JOON: Okay, I will add this change in the next PR.

mesh/mesh_initializer.py
Rahul-JOON: The names are truly subjective, I can change it to Mesh if that's better.

mesh/mesh_initializer.py
armantekinalp: I agree with you @bhosale2 . We will take care already published 'pyelastica' code base later but new code can have type hints.  @Rahul-JOON can you add type hints for the new code in this PR.

Rahul-JOON: Yes, sure I will add them.

mesh/mesh_initializer.py
```diff
@@ -0,0 +1,132 @@
...
+
+class Mesh():
+    """
+    This Mesh Initializer class uses pyvista to import mesh files in the
+    STL or OBJ file formats and initializes the necessary mesh information.
+
+    How to initialize a mesh?
+    -------------------------
+
+    mesh = MeshInitialize(r"<filepath>")
```
bhosale2, line 15: correct the docs as well, to be consistent with class name change

mesh/mesh_initializer.py
```diff
@@ -0,0 +1,132 @@
...
+
+    mesh.mesh_scale:
+        - Stores the 3 dimensions of the smallest box that could fit the mesh.
+        - Dimension: (3 spatial lengths)
+
+    mesh.mesh_center:
+        - Stores the coordinates of the position vector of the center of the smallest box that could fit the mesh.
+        - Dimension: (3 spatial coordinates)
+    """
+    def __init__(self, filepath: str):
```
bhosale2, line 45:
```suggestion
    def __init__(self, filepath: str) -> None:
```

Ali-7800 (approved):
LGTM

bhosale2 (approved):
LGTM

PR #277, opened 2023-07-14
Mesh Methods + last attribute

Rahul-JOON:
This PR adds the remaining methods and attributes necessary for mesh initialization mentioned in #270, namely:

Attributes:
     1. Orientation

Methods:
     1. Visualize
     2. Translate
     3. Scale
     4. Rotate
     5. Update


tests/test_mesh_methods.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+"""def test_visualize():
+    This function tests the visualization of the mesh.
+    Running this test will open a new window with the mesh visualization.
+    mockmesh = cube_mesh_init()
+    try:
+        mockmesh.visualize()
+    except:
+        raise RuntimeError("Visualization failed")"""
```
bhosale2, line 36: why this commented?

bhosale2 (changes requested):
Minor stuff

mesh/mesh_initializer.py
```diff
@@ -41,9 +41,53 @@ class Mesh():
...
+
+    def mesh_update(self) -> None:
+        """
+        This method updates/refreshes the mesh attributes in pyelastica geometry.
+        This needs to be performed at the first initialization as well as
+        after every method that might change the mesh attributes.
+        """
         self.mesh_center = self.mesh.center
         self.pyvista_face_normals = self.mesh.face_normals
         self.pyvista_faces = self.mesh.faces
```
armantekinalp, line 93: why do we need a variable called `pyvista_*`

armantekinalp (changes requested):

mesh/mesh_initializer.py
```diff
@@ -130,3 +175,56 @@ def mesh_scale_calculation(self, bounds: np.ndarray) -> np.ndarray:
...
+        The dummy cube's orthogonal face normals are extracted to keep record of the orientation of cube;
+        the cube is rotated with the mesh everytime the mesh.rotate() is called, so the orientation
+        of the cube is same as the mesh.
+        Lastly we find the orientation of the mesh by extracting unit face normal vector for each axis.
+        """
+        orientation = np.zeros((3, 3))
+        axis = 0
+        for i in range(1, 6, 2):
+            orientation[axis] = cube_face_normals[i]
+            axis += 1
```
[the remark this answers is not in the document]
Rahul-JOON: No, In Face Pyvista doesn't have any orientation tracking attribute right out of the box.

tests/test_mesh_methods.py
Rahul-JOON: Running this test actually opens up a pyvista plotter window mid-tests and it has to be closed manually to proceed further; I am doubtful how this would operate with github's automated tests.

mesh/mesh_initializer.py
Rahul-JOON: These variables are required to calculate the desired attributes; though separate variable assigning is not necessary.
Should I add the arguments in object-orientated mod/without creating variables?

armantekinalp: Yes I think object oriented way is better

Rahul-JOON: Okay, I will implement this.

tests/test_mesh_methods.py
```diff
@@ -0,0 +1,227 @@
...
+    assert_allclose(mockmesh.bounds, target_bounds)
+    assert_allclose(mockmesh.face_centers, target_face_centers, atol=1e-6)
+
+
+def test_mesh_scale():
+    """
+    This function tests the scaling of the mesh.
+    """
+    mockmesh = cube_mesh_init()
+    scaling_factor = np.array([2, 2, 2])
```
Ali-7800, line 126: What happens when you give it a negative or zero scale value?

tests/test_mesh_methods.py
```diff
@@ -0,0 +1,227 @@
...
+    assert_allclose(mockmesh.face_centers, target_face_centers, atol=1e-6)
+
+
+def test_mesh_rotate():
+    """
+    This function tests the rotation of the mesh.
+    """
+    mockmesh = cube_mesh_init()
+    rotation_angle = 90.0
+    rotation_axis = np.array([1, 0, 0])
```
Ali-7800, line 202: What happens when you give it a non-unit vector? does it just use the direction of the vector? or does it normalize and scale the rotation angle by the magnitude?

tests/test_mesh_methods.py
Rahul-JOON: For zero scale value, the mesh becomes not existent.
And for negative scale value, the mesh actually reverses in the respective axis and is then scaled by the factor.

Both the cases can be imagined for a uniform shape but as the shapes get complex, this gets hard to analyze, although one can test the effects by visualizing the mesh.
Even multiplication by a factor of -1 could be useful to reverse the mesh, then again if different scaling factors are given for different axis, the shape and connectivity of the mesh might suffer.

tests/test_mesh_methods.py
Rahul-JOON: No, it just uses the direction of the vector to form the rotation axis.

Ali-7800 (approved):
LGTM

armantekinalp (approved):
LGTM

bhosale2 (approved):
LGTM

PR #293, opened 2023-08-21
Surface subclass: plane, and tests

Ali-7800:
Third contact module PR mentioned in [Issue](https://github.com/GazzolaLab/PyElastica/issues/113) 

This PR includes:
- New plane surface subclass, will be used to define rod-plane and cylinder-plane contact classes that will replace `InteractionPlane`, `AnistropicFrictionalPlane` and `InteractionPlaneRigidBody` in interaction.
- Plane object initialization test.

### DIFF
--- a/elastica/surface/plane.py
+++ b/elastica/surface/plane.py
@@ -7,7 +7,7 @@ from elastica.utils import Tolerance
 
 
 class Plane(SurfaceBase):
-    def __init__(self, plane_origin, plane_normal):
+    def __init__(self, plane_origin: np.ndarray, plane_normal: np.ndarray):
         """
         Plane surface initializer.
 
@@ -29,4 +29,3 @@ class Plane(SurfaceBase):
         )
         self.normal = plane_normal.reshape(3)
         self.origin = plane_origin.reshape(3, 1)
-        self.tol = 1e-4
--- a/elastica/surface/plane.py
+++ b/elastica/surface/plane.py
@@ -27,5 +27,5 @@ class Plane(SurfaceBase):
             atol=Tolerance.atol(),
             err_msg="plane normal is not a unit vector",
         )
-        self.normal = plane_normal.reshape(3)
+        self.normal = np.asarray(plane_normal).reshape(3)
         self.origin = plane_origin.reshape(3, 1)
--- a/elastica/surface/plane.py
+++ b/elastica/surface/plane.py
@@ -13,12 +13,12 @@ class Plane(SurfaceBase):
 
         Parameters
         ----------
-        plane_origin: numpy.ndarray
-        2D (dim, 1) array containing data with 'float' type.
-        Origin of the plane.
-        plane_normal: numpy.ndarray
-        2D (dim, 1) array containing data with 'float' type.
-        The normal vector of the plane, must be normalized.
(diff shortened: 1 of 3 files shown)


elastica/surface/plane.py
```diff
@@ -0,0 +1,32 @@
...
+        2D (dim, 1) array containing data with 'float' type.
+        The normal vector of the plane, must be normalized.
+        """
+
+        assert_allclose(
+            np.linalg.norm(plane_normal),
+            1,
+            atol=Tolerance.atol(),
+            err_msg="plane normal is not a unit vector",
+        )
```
armantekinalp, line 29: Can you add a test to check error message

armantekinalp (changes requested):
Minor comment

elastica/surface/plane.py
Ali-7800: done

armantekinalp (approved):
LGTM

elastica/surface/plane.py
```diff
@@ -0,0 +1,32 @@
...
+        self.normal = plane_normal.reshape(3)
+        self.origin = plane_origin.reshape(3, 1)
+        self.tol = 1e-4
```
bhosale2, line 32: why 1e-4?

tests/test_surface/test_plane.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+    with pytest.raises(AssertionError) as excinfo:
+        test_plane = Plane(plane_origin, invalid_plane_normal)
+    assert "plane normal is not a unit vector" in str(excinfo.value)
+
+
+if __name__ == "__main__":
+    from pytest import main
+
+    main([__file__])
```
bhosale2, line 47: not needed
```suggestion
```

bhosale2 (changes requested):

elastica/surface/plane.py
bhosale2: add a one lin explanation if possible

Ali-7800: It is the penetration tolerance between the plane and the rod-like object, I got it from InteractionPlane, maybe I should put it in the contact class instead of here.

bhosale2 (approved):
LGTM

elastica/surface/plane.py
```diff
@@ -0,0 +1,31 @@
...
+        The normal vector of the plane, must be normalized.
+        """
+
+        assert_allclose(
+            np.linalg.norm(plane_normal),
+            1,
+            atol=Tolerance.atol(),
+            err_msg="plane normal is not a unit vector",
+        )
+        self.normal = plane_normal.reshape(3)
```
skim0119, line 30:
```suggestion
        self.normal = np.asarray(plane_normal).reshape(3)
```

elastica/surface/plane.py
```diff
@@ -0,0 +1,31 @@
...
+        self.origin = plane_origin.reshape(3, 1)
```
skim0119, line 31:
```suggestion
        self.origin = np.asarray(plane_origin).reshape(3, 1)
```

skim0119 (approved):

PR #298, opened 2023-09-09
Rod sphere contact

Rahul-JOON:
This PR attempts to add Rod SPhere contact functionality as per #266.

PR #301, opened 2023-10-02
266 rod and sphere contact temp

Ali-7800:
Finishing requested changes in [#298 ](https://github.com/GazzolaLab/PyElastica/pull/298)

### DIFF
--- a/elastica/contact_forces.py
+++ b/elastica/contact_forces.py
@@ -2,13 +2,14 @@ __doc__ = """ Numba implementation module containing contact between rods and ri
 
 from elastica.typing import RodType, SystemType, AllowedContactType
 from elastica.rod import RodBase
-from elastica.rigidbody import Cylinder
+from elastica.rigidbody import Cylinder, Sphere
 from elastica.contact_utils import (
     _dot_product,
     _norm,
     _find_min_dist,
     _prune_using_aabbs_rod_cylinder,
     _prune_using_aabbs_rod_rod,
+    _prune_using_aabbs_rod_sphere,
 )
 
 from elastica._linalg import _batch_product_k_ik_to_ik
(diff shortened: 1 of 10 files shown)


armantekinalp (approved):
LGTM

elastica/contact_forces.py
```diff
@@ -725,3 +861,120 @@ def apply_contact(
...
+        ):
+            raise TypeError(
+                "Systems provided to the contact class have incorrect order/type. \n"
+                " First system is {0} and second system is {1}. \n"
+                " First system should be a rod, second should be a sphere".format(
+                    system_one.__class__, system_two.__class__
+                )
+            )
+
+    def apply_contact(self, system_one: RodType, system_two: SystemType):
```
armantekinalp, line 939: Some of the other contact methods have `*args` and `*kwargs` . I think you are passing time through them, is this correct? If that is the case why not passing time explicitly. Why do we need `*args` and `*kwargs`. Also sphere contact doesn't have args and kwargs.

armantekinalp (commented):
RodSphereContact import was missing in the __init__.py I fixed it just update your branch.

elastica/contact_forces.py
```diff
@@ -423,6 +424,143 @@ def _calculate_contact_forces_self_rod(
...
+            external_forces_rod[..., i]
+            + external_forces_rod[..., i + 1]
+            + internal_forces_rod[..., i]
+            + internal_forces_rod[..., i + 1]
+        )
+        equilibrium_forces = -rod_elemental_forces + external_forces_sphere[..., 0]
+
+        normal_force = _dot_product(equilibrium_forces, distance_vector)
+        # Following line same as np.where(normal_force < 0.0, -normal_force, 0.0)
+        normal_force = abs(min(normal_force, 0.0))
```
armantekinalp, line 485: Normal force is not used, lets remove these lines. Can you also check other classes as well?

elastica/contact_forces.py
Ali-7800: Removed `*args` and `**kwargs` from `apply_contact()` as they are not necessary

elastica/contact_forces.py
Ali-7800: removed from rod-sphere and rod-cylinder, need to remove from rod-rod contact and to retune the rod-rod examples accordingly

elastica/contact_forces.py
skim0119: We should keep `*args` and `*kwargs` like `forcing`, `damping`, or `constraints`. It is an interface for extra user parameters. I'm unsure if anyone will ever use it, but the feature is there for other modules.

We are not passing time, because we don't have any `ramping` behavior for contact, but it might be useful in the future.

@armantekinalp Any thoughts?

skim0119 (approved):
Looks good. One comment for removing `args` and `kwargs` for this module.

armantekinalp:
@skim0119 but user cannot control those `*args` and `*kwargs`, they are different then the ones we use while constructing the class . So I don't think we need. We can pass time explicitly? What do you think

skim0119:
> @skim0119 but user cannot control those `*args` and `*kwargs`, they are different then the ones we use while constructing the class . So I don't think we need. We can pass time explicitly? What do you think

I agree that we don't need, but I was thinking that its a feature that doesn't need to be removed. Maybe lets remove it, merge PR as it is, but consider removing args and kwargs in other modules in the future? Its probably better to pass parameters more explicitly.

elastica/contact_forces.py
```diff
@@ -604,10 +719,7 @@ def _check_systems_validity(
                 )
             )
 
-    def apply_contact(
-        self, system_one: RodType, system_two: SystemType, *args, **kwargs
-    ):
-
+    def apply_contact(self, system_one: RodType, system_two: SystemType):
```
armantekinalp, line 722:
```suggestion
    def apply_contact(self, system_one: RodType, system_two: SystemType) -> None:
```

elastica/contact_forces.py
```diff
@@ -485,7 +600,7 @@ def _check_systems_validity(
                 "If you want self contact, use RodSelfContact instead"
             )
 
-    def apply_contact(self, system_one: RodType, system_two: RodType, *args, **kwargs):
+    def apply_contact(self, system_one: RodType, system_two: RodType):
```
armantekinalp, line 603:
```suggestion
    def apply_contact(self, system_one: RodType, system_two: RodType) -> None:
```

elastica/contact_forces.py
```diff
@@ -725,3 +834,120 @@ def apply_contact(
...
+        ):
+            raise TypeError(
+                "Systems provided to the contact class have incorrect order/type. \n"
+                " First system is {0} and second system is {1}. \n"
+                " First system should be a rod, second should be a sphere".format(
+                    system_one.__class__, system_two.__class__
+                )
+            )
+
+    def apply_contact(self, system_one: RodType, system_two: SystemType):
```
armantekinalp, line 912:
```suggestion
    def apply_contact(self, system_one: RodType, system_two: SystemType) -> None:
```

armantekinalp (approved):
LGTM three suggestions. You can directly commit

PR #314, opened 2023-10-27
RodPlaneContact and RodPlaneContactWithAnisotropicFriction Classes

Ali-7800:
Fourth contact module PR mentioned in [Issue](https://github.com/GazzolaLab/PyElastica/issues/113) 
This PR includes:
- Added RodPlaneContact, and RodPlaneContactWithFriction which are based on InteractionPlane and AnisotropicFrictionalPlane from interaction.
- Updated base_system to call _call_contacts last (needed so the external forces are applied before contact)
- Updated memory_block to accept surface class when appended to the simulation

### DIFF
--- a/elastica/contact_forces.py
+++ b/elastica/contact_forces.py
@@ -3,6 +3,7 @@ __doc__ = """ Numba implementation module containing contact between rods and ri
 from elastica.typing import RodType, SystemType, AllowedContactType
 from elastica.rod import RodBase
 from elastica.rigidbody import Cylinder, Sphere
+from elastica.surface import Plane
 from elastica.contact_utils import (
     _dot_product,
     _norm,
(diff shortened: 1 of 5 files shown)


elastica/contact_forces.py
sy-cui, line 1: Are these duplicate functions from the `interaction.py` file?

elastica/contact_utils.py
sy-cui, line 1: Same here?

sy-cui (commented):
Seems like the additions are the functions from the interaction file (unless there are minor differences that I didn't catch?). Should the old ones be removed if we are moving them here?

elastica/contact_forces.py
Ali-7800: yes, the plan to refactor the stuff from `interaction.py` in the next PR, all the duplicate functions will be refactored similar to what was done to `node_to_element_pos_or_vel` in `interaction.py` before.

elastica/contact_utils.py
Ali-7800: Same as above comment

sy-cui (approved):
LGTM. The contact forces file seems a bit overcrowded. Maybe in a future update separate them into different locations.

armantekinalp (approved):
LGTM. @Ali-7800 please also start refactoring PR for interaction right after this PR.

Issue #255, opened 2023-05-13
Type hints and static type checking for PyElastica


bhosale2:
We should have type hints and static type checking throughout PyElastica using `mypy`, which adds more context to function arguments and reduces the chances of error while coding.

ankith26:
Hello! I'd like to work on this issue if @armantekinalp hasn't started work on this already, and wants me to instead.

armantekinalp:
sure go ahead

PR #341, opened 2024-03-09
Typing: Add typehinting for `Rod` modules

ankith26:
This PR does some initial work on https://github.com/GazzolaLab/PyElastica/issues/255

This is a WIP, and I've opened this PR to get some feedback, before I go ahead and look into more modules.

### DIFF
--- a/elastica/rod/knot_theory.py
+++ b/elastica/rod/knot_theory.py
@@ -39,7 +39,7 @@ class KnotTheoryCompatibleProtocol(Protocol):
     def base_length(self) -> np.ndarray: ...
 
 
-class KnotTheory(KnotTheoryCompatibleProtocol):
+class KnotTheory:
     """
     This mixin should be used in RodBase-derived class that satisfies KnotCompatibleProtocol.
     The theory behind this module is based on  the method from Klenin & Langowski 2000 paper.
--- a/elastica/rod/knot_theory.py
+++ b/elastica/rod/knot_theory.py
@@ -14,6 +14,7 @@ from typing import Protocol, Union
 
 from numba import njit
 import numpy as np
+from numpy.typing import NDArray
 
 from elastica.rod.rod_base import RodBase
 from elastica._linalg import _batch_norm, _batch_dot, _batch_cross
@@ -38,7 +39,7 @@ class KnotTheoryCompatibleProtocol(Protocol):
     def base_length(self) -> np.ndarray: ...
 
 
-class KnotTheory:
+class KnotTheory(KnotTheoryCompatibleProtocol):
     """
     This mixin should be used in RodBase-derived class that satisfies KnotCompatibleProtocol.
     The theory behind this module is based on  the method from Klenin & Langowski 2000 paper.
@@ -46,7 +47,7 @@ class KnotTheory:
     KnotTheory can be mixed with any rod-class based on RodBase::
 
         class MyRod(RodBase, KnotTheory):
-            def __init__(self):
+            def __init__(self) -> None:
                 super().__init__()
         rod = MyRod(...)
 
@@ -78,7 +79,7 @@ class KnotTheory:
 
     MIXIN_PROTOCOL = Union[RodBase, KnotTheoryCompatibleProtocol]
 
-    def compute_twist(self: MIXIN_PROTOCOL):
+    def compute_twist(self: MIXIN_PROTOCOL) -> NDArray[np.floating]:
         """
         See :ref:`api/rods:Knot Theory (Mixin)` for the detail.
         """
@@ -91,8 +92,8 @@ class KnotTheory:
     def compute_writhe(
         self: MIXIN_PROTOCOL,
         type_of_additional_segment: str = "next_tangent",
-        alpha: float = 1.0,
-    ):
+        alpha: np.floating = 1.0,
+    ) -> NDArray[np.floating]:
         """
         See :ref:`api/rods:Knot Theory (Mixin)` for the detail.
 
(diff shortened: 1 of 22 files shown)


skim0119:
@ankith26 I allowed the `github-action` to run for you to check. For formatting, you can use `make formatting` or setup-up a `pre-commit hook`.

skim0119:
@armantekinalp Since we are bumping up to `3.10`, lets see if we can pass this PR and add `mypy` checker.

armantekinalp:
@ankith26 if you are still interested in this issue can you resolve the conflicts please.

ankith26:
I have force pushed with the conflicts fixed. The conflicts were mainly due to the fact that I was adding type hints to the functions that were being deprecated in the `update-0.3.3` branch. I removed my changes in these cases

armantekinalp:
@ankith26 thanks. Can you also see why the CI builds fail?

ankith26:
While resolving the conflicts I forgot to remove an import that is now unneeded. I have removed that import to fix the fail, and also made sure `make autoflake8-check` passed locally

skim0119:
@ankith26 Hi -- I added `mypy` on GitHub action, and, as we have somewhat expected, it is becoming a large job to finish. We have decided that we should split the work on #255, and we can probably pass this PR without completely checking out all CIs.

Thank you again for your contribution. If you can finish typehinting for `elastica/rod` directory, we'll merge this PR with #367. You can run `mypy --config-file pyproject.toml elastica/rod`

ankith26:
> If you can finish typehinting for elastica/rod directory, we'll merge this PR with https://github.com/GazzolaLab/PyElastica/pull/367.

Sure I'd like to work on this this

ankith26:
Update: I couldn't do much on this PR in the past week due to being busy with other commitments, but I plan to get this PR done from my side in the coming week.

armantekinalp:
@ankith26 no worries, also please see the other open PRs related to mypy. That could be useful for you.

ankith26:
I have implemented a workaround for the float issue. In `elastica.typing` I have defined `Float: TypeAlias = Union[np.floating, float]`, and now this new type can be used where a single float is expected

skim0119:
> I have implemented a workaround for the float issue. In `elastica.typing` I have defined `Float: TypeAlias = Union[np.floating, float]`, and now this new type can be used where a single float is expected

In this PR, please just use `np.floating` whenever you need to type `float` type for now, and don't create another type alias for now. We'll make that decision later.

> mypy also doesn't like the usage of numba decorators, as they apparently don't preserve typing info.

The function should be typed as if they are just `python`/`numpy` function. I understand that 
`numba` decorator will change the function call so `mypy` would raise error. For now, just ignore them.

ankith26:
I have done some more incremental progress, and have removed the `Float` type as you asked.

armantekinalp:
Lets keep in draft until it is ready. Since it burns CI hours for every new commit

skim0119:
Thank you for your hard work. There are some inconsistency because the original code was only developed for `Rod` implementation, and later we decided to add other types like `RigidBody` and `Surfaces`. I'll take care of organizing properties in other PR.

To answer your question:

> 1. The following attributes are expected in a `SystemType` instance in various places across the codebase, but not typed into `SystemType`. How should these be dealt with?

`SystemType` should include all `Rod`, `RigidBody`, and `Surface`, so please use it whenever these objects needs to be passed or returned. For now, don't worry about any `mypy` error caused by missing property.

> 2. Similarly, `RodType` is used in a couple of places where these attributes are expected:

Same as above. We'll have `SystemType` which is union of `RodType`, `RigidBodyType`, and `SurfaceType`. If you need to type-hint `Rod` specific module, use `RodType`.

> 3. `RodType` incompatible with `SystemType` error, what is the actual relation between the two and how can this be fixed in typing?

Don't worry about it for now.

> 4. `ConstraintBase`: Why are properties `constrained_position_idx` and `constrained_director_idx` typed with optional?

There are some boundary condition implementations that do not use `_idx` in the definition. You don't need to modify at this point.

> 5. How to type a `MemoryBlock` in `cosserat_rot.py`

Use `RodType`.

> 6. `KnotTheoryCompatibleProtocol` defines `director_collection` and `radius` read-only, but value set in `cosserat_rot.py`

The protocol for `KnotTheory` shouldn't write to those properties. If you see any error, I would assume it is because `RodType` is not fully defined yet.

> 7. `CosseratRod` instance expects the following attributes, but these aren't set in the constructor:

You don't need to worry about it, it will be included in `RodType` in the later commit.

ankith26:
This PR is now done from my side, it is ready for review.

elastica/rod/knot_theory.py
```diff
@@ -38,15 +39,15 @@ def radius(self) -> np.ndarray: ...
     def base_length(self) -> np.ndarray: ...
 
 
-class KnotTheory:
+class KnotTheory(KnotTheoryCompatibleProtocol):
```
skim0119, line 42:
```suggestion
class KnotTheory:
```

skim0119: No need to inherit protocol here. Also because of this, unit-tests are not passing.

skim0119 (changes requested):

skim0119:
Looks good. One minor change, and we can probably merge.

ankith26:
I had added that bit to fix some mypy issues, but didn't realise it broke the unit tests. I have reverted that change, as suggested.

skim0119 (approved):

PR #342, opened 2024-03-10
Add support for python 3.11 and 3.12

ankith26:
I'm taking a more conservative approach in this PR, where I'm only bumping the dependencies in the new python version case, but I am open to implementing other approaches.


armantekinalp:
I am closing this PR, since python 3.12 is released soon. We consider this in next version. Thank you for contributing.

PR #363, opened 2024-04-23
update: python version of pyelastica and its CI workflow.

armantekinalp:
Setup tools that GitHub CI uses depreciated Python 3.8 and 3.9 for latest macos, so we are changing the python version of pyelastica.


skim0119 (approved):
LGTM!

PR #368, opened 2024-04-28
Black 24.3.0 formatting

skim0119:
We have some security issue #350 that we've been declining. This should fix any formatting issues for new `black-24.3.0`.

@armantekinalp I already upgraded the black version in `update-0.3.3`.


armantekinalp (approved):
LGTM

PR #372, opened 2024-04-29
Typing: timestepper module

skim0119:
# Type-hinting `timestepper` module

```sh
-> % mypy --config-file pyproject.toml elastica/timestepper
Success: no issues found in 5 source files
```

- Partially address #255

## New Protocols

- `elastica/system/protocol.py`: generic protocol for all `SystemType`
  - Also includes `SymplecticSystemProtocol` and `ExplicitSystemProtocol`
- `elastica/timestepper/protocol.py`:
  - generic protocol for any `StepperProtocol`
    - Extended protocols: `StatefulStepperProtocol`
  - `MethodCollectorProtocol` for class that generates operators.
  - `MemoryProtocol` for state representation of timesteppers. (unused) 

## Refactoring

- Unused `__init__`, `args`, and `kwargs` are removed.
- `SymplecticStepperTag` and `ExplicitStepperTag` are now in `tag.py`
  - All stepper tagging are now done using `elastica.timestepper.tag.tag`.
- In testing, some `MockRod` and `MockRigidBody` was created using `TypeAlias`, which is not an intended usage. It is refactored to import original base classes.

## Minor changes
- `isinstance` is safer type-comparison instead of `type(...) == type(...)`
- Change some `float` types to `np.floating`
  - Not sure yet this is a good idea. Maybe we should make generic datatype: `FLOAT: TypeAlias = float | np.floating`.
- Some commented-out by string is replaced with proper python-commenting for better syntax highlighitng.


elastica/modules/base_system.py
```diff
@@ -181,22 +183,22 @@ def finalize(self):
                 self._feature_group_synchronize.pop(index)
             )
 
-    def synchronize(self, time: float):
+    def synchronize(self, time: np.floating):
```
armantekinalp, line 186: why np.floating? instead of float. Also there are np.float64 or np.float32, is np.floating more general

skim0119: np.floating includes all size of float. Python float and np.float ling is different, and I don't have exact justification other than numpy has numerically accurate representation where python-float has better print

elastica/timestepper/symplectic_steppers.py
```diff
@@ -29,16 +32,20 @@
...
+    ) -> np.floating:
         for kin_prefactor, kin_step, dyn_step in _steps_and_prefactors[:-1]:
             kin_step(TimeStepper, System, time, dt)
             time += kin_prefactor(TimeStepper, dt)
             System.update_internal_forces_and_torques(time)
             dyn_step(TimeStepper, System, time, dt)
 
         # Peel the last kinematic step and prefactor alone
-        last_kin_prefactor = _steps_and_prefactors[-1][0]
+        last_kin_prefactor: Callable[..., np.floating] = _steps_and_prefactors[-1][0]
```
armantekinalp, line 48:
```suggestion
        last_kin_prefactor: OperatorType = _steps_and_prefactors[-1][0]
```

armantekinalp (approved):
LGTM nice work @skim0119

PR #373, opened 2024-05-02
Typing: `timestepper` fix pytest and cleanup integration routine

skim0119:
# Type-hinting and cleanup `timestepper` routine

```sh
-> % mypy --config-file pyproject.toml elastica/timestepper
Success: no issues found in 5 source files
```

```sh
-> % make test
2181 passed, 1 skipped, 30 xfailed, 16 xpassed, 4 warnings in 16.18s
```

- Partially address #255

## Modification on user-API

- Integrator routine is simplified:

```py
timestepper: SymplecticStepperProtocol = PositionVerlet()
dt = 1e-3
for _ in range(100):  # 100 timestepping
  timestepper.step(simulator, time=0.0, dt=dt)
```

- Previous `extend_stepper_interface` and `integrate` should work same as before.
  - Added deprecation tag on `extend_stepper_interface`.
  - People should be able to just use `integrate` or custom stepping as above.

## Minor changes

- Updated and fixed any `mypy` type-hinting issues
- `1_Timoshenko_Beam.ipynb` is updated to use new time-step routine instead of `extend_stepper_interface` method.
- `timestepper.tag` removed as we no-longer need to check type of the stepping method.
- Fixed testing
- `autoflake8` -> `autoflake` : The forked repo `autoflake8` is now deprecated, and `autoflake` is continued to be maintained.


elastica/timestepper/symplectic_steppers.py
```diff
@@ -120,88 +135,46 @@ def do_step(
 
         return time
 
+    def step_single_instance(
+        self: SymplecticStepperProtocol,
+        System: SystemType,
+        time: np.floating,
+        dt: np.floating,
+    ) -> np.floating:
```
armantekinalp, line 143: Why do we need this?

elastica/timestepper/__init__.py
```diff
@@ -10,57 +10,52 @@
...
-    StatefulStepper: StatefulStepperProtocol,
-    System: SystemType,
+    stepper: StepperProtocol,
+    systems: SystemCollectionType,
+    final_time: float,
+    n_steps: int,
+    restart_time: float,
+    progress_bar: bool,
+) -> float: ...
+
```
armantekinalp, line 54: why do we need two of these?

elastica/timestepper/__init__.py
```diff
@@ -86,17 +81,15 @@ def integrate(
...
+    time = np.float_(restart_time)
 
-    for i in tqdm(range(n_steps), disable=(not progress_bar)):
-        time = do_step(StatefulStepper, stages_and_updates, System, time, dt)
+    if is_system_a_collection(systems):
+        for i in tqdm(range(n_steps), disable=(not progress_bar)):
+            time = stepper.step(systems, time, dt)  # type: ignore[arg-type]
+    else:
+        for i in tqdm(range(n_steps), disable=(not progress_bar)):
+            time = stepper.step_single_instance(systems, time, dt)  # type: ignore[arg-type]
```
armantekinalp, line 92: Why do we need this step_single_instance call. Is this for tests?

armantekinalp: Seems like this is needed for Explicit steppers can you confirm.?

armantekinalp (changes requested):
I have couple of questions. Please address them. Overall looks good. Thanks @skim0119

elastica/timestepper/symplectic_steppers.py
skim0119: This is for testing. We have test cases that test stepping with single objects. I think it would be good idea to keep them in place.

elastica/timestepper/__init__.py
skim0119: `step_single_instance` is mostly used for testing at this point.

armantekinalp (approved):
LGTM

PR #378, opened 2024-05-07
Refactor: contact synchronization order

skim0119:
## Minor change
- Refactor code to make sure the contact-synchronization is organized in its own feature iterator.
- Cleaning syntax

### DIFF
--- a/elastica/modules/base_system.py
+++ b/elastica/modules/base_system.py
@@ -46,6 +46,7 @@ class BaseSystemCollection(MutableSequence):
         # Each component (Forcing, Connection, etc.) registers the executable (callable) function
         # in the group that that needs to be executed. These should be initialized before mixin.
         self._feature_group_synchronize: Iterable[Callable[[float], None]] = []
+        self._feature_group_synchronize_contact: Iterable[Callable[[float], None]] = []
         self._feature_group_constrain_values: Iterable[Callable[[float], None]] = []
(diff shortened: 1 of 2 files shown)


elastica/modules/base_system.py
```diff
@@ -169,22 +170,13 @@ def finalize(self):
...
-            self._feature_group_synchronize.append(
-                self._feature_group_synchronize.pop(index)
-            )
 
     def synchronize(self, time: float):
         # Collection call _feature_group_synchronize
         for feature in self._feature_group_synchronize:
             feature(time)
+        for feature in self._feature_group_synchronize_contact:
+            feature(time)
```
armantekinalp, line 179: Why do we need nested loops, and not add end of `self._feature_group_synchronize` list?

armantekinalp: yeah I am just saying why not do something like this somewhere in line 172. 

```python
self._feature_group_synchronize = self._self._feature_group_synchronize_connections + self._feature_group_synchronize_forcing  +  self._feature_group_synchronize_contact

```

We can explicitly set the order for forcing, connections and contact. Now connections are also part of `self._feature_group_synchronize`

armantekinalp: This is actually something I wanted to do, since we don't have a protocol in calling order of connections, forcing and contact before. We were only calling contact at the end.

PR #379, opened 2024-05-09
Explicit operation order determined by user definition

skim0119:
# Explicit operation order

- User-API would remain the same

## Description

Previous implementation has ambiguity in determining the order of operation for different boundary conditions: `forcing`, `contact`, `friction`, `connection`, etc. For example, `contact` and `friction` must be calculated after all other boundary conditions are in place, and strictly `contact` must be computed before `friction` since `friction` depends on `contact`. In various place in the code, we tried to rearange the order without explicitly letting user know: https://github.com/GazzolaLab/PyElastica/blob/87a2e74eb5cd1de0671847753efb630afaa1acf5/elastica/modules/base_system.py#L173-L182, https://github.com/GazzolaLab/PyElastica/blob/87a2e74eb5cd1de0671847753efb630afaa1acf5/elastica/modules/forcing.py#L72-L81.
This cause ambiguity on user, especially when user wants to define their own boundary condition.

## Solution

As long as we let users to write their own boundary conditions, we cannot anticipate every possible cases of operation dependencies. We are shifting to the policy that user must be aware of the boundary operation order.

This PR removes any "back-end" sorting the operation order. __Operation order is now strictly defined by the order of user inserting any boundary condition with `.using`.__ This applies to `forcing`, `contact`, and `connection` as of now.

## Some remaining TODOs:

- [x] Add warning messages on packaged `contact` and `friction` modules
- [x] Refactor synchronize operations: "first-in" boundary condition computes first
- [x] Test
  - [x] Fix existing tests
  - [x] Implement new test for testing operation order
- [x] Documentation ([f0fc544](https://github.com/GazzolaLab/PyElastica/pull/379/commits/f0fc544bb8be1458b03b41a5458ff453b04fea33))

### DIFF
--- a/elastica/modules/connections.py
+++ b/elastica/modules/connections.py
@@ -5,7 +5,6 @@ Connect
 Provides the connections interface to connect entities (rods,
 rigid bodies) using joints (see `joints.py`).
 """
-import functools
 import numpy as np
 from elastica.joint import FreeJoint
 
@@ -60,16 +59,15 @@ class Connections:
         sys_dofs = [self._systems[idx].n_elems for idx in sys_idx]
 
         # Create _Connect object, cache it and return to user
-        _connector = _Connect(*sys_idx, *sys_dofs)
-        _connector.set_index(first_connect_idx, second_connect_idx)
-        self._connections.append(_connector)
-        self._feature_group_synchronize.append_id(_connector)
+        _connect = _Connect(*sys_idx, *sys_dofs)
+        _connect.set_index(first_connect_idx, second_connect_idx)
+        self._connections.append(_connect)
+        self._feature_group_synchronize.append_id(_connect)
 
-        return _connector
+        return _connect
 
     def _finalize_connections(self):
         # From stored _Connect objects, instantiate the joints and store it
-
         # dev : the first indices stores the
         # (first rod index, second_rod_idx, connection_idx_on_first_rod, connection_idx_on_second_rod)
         # to apply the connections to.
@@ -82,8 +80,7 @@ class Connections:
 
             # FIXME: lambda t is included because OperatorType takes time as an argument
             def apply_forces(time):
-                return functools.partial(
-                    connect_instance.apply_forces,
+                connect_instance.apply_forces(
                     system_one=self._systems[first_sys_idx],
                     index_one=first_connect_idx,
                     system_two=self._systems[second_sys_idx],
@@ -91,8 +88,7 @@ class Connections:
                 )
 
             def apply_torques(time):
-                return functools.partial(
-                    connect_instance.apply_torques,
+                connect_instance.apply_torques(
                     system_one=self._systems[first_sys_idx],
                     index_one=first_connect_idx,
                     system_two=self._systems[second_sys_idx],
@@ -102,6 +98,9 @@ class Connections:
             self._feature_group_synchronize.add_operators(
                 connection, [apply_forces, apply_torques]
             )
+
+        self._connections = []
+        del self._connections
 
         # Need to finally solve CPP here, if we are doing things properly
         # This is to optimize the call tree for better memory accesses
@@ -156,7 +155,7 @@ class _Connect:
     def set_index(self, first_idx, second_idx):
         # TODO assert range
         # First check if the types of first rod idx and second rod idx variable are same.
-        assert type(first_idx) == type(
+        assert type(first_idx) is type(
             second_idx
         ), "Type of first_connect_idx :{}".format(
             type(first_idx)
--- a/elastica/modules/connections.py
+++ b/elastica/modules/connections.py
@@ -99,12 +99,17 @@ class Connections:
                 connection, [apply_forces, apply_torques]
             )
 
+            self.warnings(connection)
+
         self._connections = []
         del self._connections
 
         # Need to finally solve CPP here, if we are doing things properly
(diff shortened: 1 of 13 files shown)


elastica/modules/forcing.py
```diff
@@ -54,37 +54,24 @@ def _finalize_forcing(self):
...
+            apply_forces = functools.partial(
+                forcing_instance.apply_forces, system=self._systems[sys_id]
+            )
+            apply_torques = functools.partial(
+                forcing_instance.apply_torques, system=self._systems[sys_id]
+            )
+
+            self._feature_group_synchronize.add_operators(
+                ext_force_torque, [apply_forces, apply_torques]
+            )
```
armantekinalp, line 71: Don't we need to pass time here?

skim0119: No, apply_force methods already include time as a parameter. Actually, we should add that parameter to connection and contact for consistency, which is why their implementation look a little different.

elastica/modules/connections.py
```diff
@@ -60,49 +59,53 @@ def connect(
...
+                connect_instance.apply_torques(
+                    system_one=self._systems[first_sys_idx],
+                    index_one=first_connect_idx,
+                    system_two=self._systems[second_sys_idx],
+                    index_two=second_connect_idx,
+                )
+
+            self._feature_group_synchronize.add_operators(
+                connection, [apply_forces, apply_torques]
+            )
```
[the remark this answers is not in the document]
armantekinalp: Or is it because of time argument?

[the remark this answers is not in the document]
skim0119: #380 I added new issue, we'll clean out this part in later PR.

elastica/modules/contact.py
```diff
@@ -61,30 +60,26 @@ def _finalize_contact(self) -> None:
...
                 self._systems[second_sys_idx],
             )
 
+            def apply_contact(time):
+                contact_instance.apply_contact(
+                    system_one=self._systems[first_sys_idx],
+                    system_two=self._systems[second_sys_idx],
+                )
+
+            self._feature_group_synchronize.add_operators(contact, [apply_contact])
```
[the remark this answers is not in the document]
skim0119: #380 I added new issue, we'll clean out this part in later PR.

elastica/modules/feature_group.py
```diff
@@ -0,0 +1,26 @@
...
+        for operator in operator_chain:
+            yield operator
+
+    def append_id(self, feature):
+        self._operator_ids.append(id(feature))
+        self._operator_collection.append([])
+
+    def add_operators(self, feature, operators: list[OperatorType]):
+        idx = self._operator_ids.index(id(feature))
+        self._operator_collection[idx].extend(operators)
```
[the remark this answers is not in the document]
skim0119: [c4ad237](https://github.com/GazzolaLab/PyElastica/pull/379/commits/c4ad237b784490611a26b62314749e56c584b0ed)
I added documentation and how to use. Also added unittest to have a script of "expected" usage.

elastica/modules/forcing.py
```diff
@@ -54,37 +57,37 @@ def _finalize_forcing(self):
...
+                external_force_and_torque, [apply_forces, apply_torques]
+            )
+
+            self.warnings(external_force_and_torque)
+
+        self._ext_forces_torques = []
+        del self._ext_forces_torques
+
+    def warnings(self, external_force_and_torque):
+        from elastica.interaction import AnisotropicFrictionalPlane
```
armantekinalp, line 82: We moved to friction plane under contact. Can you change accordingly.

elastica/modules/operator_group.py
```diff
@@ -0,0 +1,62 @@
+from elastica.typing import OperatorType
+
+from collections.abc import Iterable
+
+import itertools
+
+
+class OperatorGroupFIFO(Iterable):
```
armantekinalp, line 8: FIFO does it mean first initialized first operated? Can we Write long maybe?

elastica/modules/contact.py
```diff
@@ -61,30 +63,36 @@ def _finalize_contact(self) -> None:
...
+
+            self.warnings(contact)
```
armantekinalp, line 83: smart way to check. Only issue, is what happens if we have rod-rod contact and frictional plane at the same time?

elastica/modules/operator_group.py
skim0119: FIFO: first in first out
I think it is common-enough terminology for queuing

elastica/modules/contact.py
skim0119: You'll get warning 😄

armantekinalp: Or two contacts.

elastica/modules/forcing.py
skim0119: I see `AnisotropicFrictionalPlane` in `interaction.py`:
https://github.com/GazzolaLab/PyElastica/blob/e71e54790d5b782ad1a2d9856888b09783251fd0/elastica/interaction.py#L158C7-L158C33

armantekinalp (approved):
LGTM

elastica/modules/forcing.py
skim0119: I removed warning for AnisotropicFrictionalPlane. @Ali-7800 will remove the class.

PR #383, opened 2024-05-13
Feature: check require-modules for each system

skim0119:
## Description

Currently, defining `SystemCollection` was independent from what kind of `system` we append, mainly because there was no dependency between the `system` (like `CosseratRod`) and `modules` (like `Constraints` and `Damping`). Some system (mainly `ring-rod`) must require `Constraints` module, which introduce dependency between system and modules. To accomodate for such dependency in generic way, I added `REQUISITE_MODULES` class variable to each system.

In the [future update](https://github.com/skim0119/PyElastica/blob/f73628e410a4b68de4a27a5f3ccaea34741c58e6/elastica/systems/protocol.py#L16), this behavior will be added as a `System` protocol.


armantekinalp (approved):
LGTM 👍

Issue #354, opened 2024-03-30
Inconsistent dimensions for damping constants in AnalyticalLinearDamper


sy-cui:
The linear viscous damping forces and torques within the linear and angular momentum equations are given as follows ([Gazzola 2018](https://royalsocietypublishing.org/doi/10.1098/rsos.171628)):
$$\mathbf{f}_v = \gamma_t \mathbf{v}, ~~~~~~~~ \mathbf{c}_L^v = \gamma_r \mathbf{\omega}_L $$
where the paper claims that the damping coefficients $\gamma_t$ and $\gamma_r$ can be set to equal. This statement is not strictly correct as they are dimensionally inconsistent:
$$[\gamma_t] = FL^{-1}T, ~~~~~~~ [\gamma_r] = FLT$$

The analytical linear damper assumes an eigenvalue problem with respect to the damping force, where the damped translational and angular velocities are respectively
$$v_{\text{damped}} = v_0 \exp \left(-\frac{\gamma_t}{m}t\right), ~~~~~~~ \omega_{\text{damped}} = \omega_0 \exp \left(-\frac{\gamma_r e}{J}t\right)$$
Current implementation assumes $\gamma = \gamma_t = \gamma_r$, and the input ``damping_constant`` corresponds to $\gamma / m$ where $m$ is the nodal / elemental mass. This includes a dimensional inconsistency and incurs problems in non-dimensional analysis. 

Two possible remedies are proposed:
1. The analytically rigorous approach is to include both $\gamma_t$ and $\gamma_r$ from user input, and compute the exponential coefficients based on these inputs and rod properties
2. Assume that $k = \frac{\gamma_t}{m} = \frac{\gamma_r e}{J}$, with $k$ having inverse time dimension ($[k]= T^{-1}$). The damper class collects only one user input $k$ instead of two, and no access to kinematic/material properties is required. 

The original implementation should be retained for the moment and possibly removed in a future update. If removal is expected, usage of the original approach should emit a deprecation warning.

sy-cui:
Resolved in #437

PR #392, opened 2024-06-12
Minor update: dissipation scaling between translation and rotation

skim0119:
#354


sy-cui:
Couple of points here:
* Static cases like axial stretching and timoshenko are okay. The analytical results are recovered
* Catenary case blows up with the current parameter. Might need to adjust the damping constant
* Snake and flagella cases are broken due to some issue in base_system (e.g. run the snake case to replicate the issue). It does not seem to be related to this PR. 
* The new implementation is okay, but it erases the original behavior and may break user code after the update is released. I thought the plan was to keep the original behavior but add a deprecation warning?

skim0119:
> Catenary case blows up with the current parameter. Might need to adjust the damping constant

@Ali-7800 Do you remember how you tuned the damping coefficient for Catenary case before?

> * Snake and flagella cases are broken due to some issue in base_system (e.g. run the snake case to replicate the issue). It does not seem to be related to this PR.

Those cases will be fixed with #367 

> * The new implementation is okay, but it erases the original behavior and may break user code after the update is released. I thought the plan was to keep the original behavior but add a deprecation warning?

We can do that if we consider this change as an `update`. I think we can make a hotfix if we consider this change a `bug fix`. @armantekinalp opinion?

PR #405, opened 2024-06-30
[Type] Check instantiation for system-collection and timestepper

### LINKED ISSUE #255: Type hints and static type checking for PyElastica

skim0119:
resolve #255 

## Update

- Add type-checking commonly used system-collection and timestepper for mypy-check.
  - https://github.com/skim0119/PyElastica/blob/3d7251deeaec0737eaf2897f47bb6ae9f17fdc55/elastica/timestepper/symplectic_steppers.py#L310-L314
  - https://github.com/skim0119/PyElastica/blob/3d7251deeaec0737eaf2897f47bb6ae9f17fdc55/elastica/modules/base_system.py#L280-L314
- Add mypy check for demonstration example cases for consistency


elastica/rod/protocol.py
```diff
@@ -6,14 +6,18 @@
 from elastica.systems.protocol import SystemProtocol, SlenderBodyGeometryProtocol
 
 
-class _RodEnergy(Protocol):
+class _CosseratRodStrainEnergy(Protocol):
```
sy-cui, line 9: Since it includes translational and rotational energy, why not just call it `_CosseratRodEnergy`

sy-cui (commented):
One small comment. Otherwise LGTM

elastica/rod/protocol.py
skim0119: Yup. [c5319a6](https://github.com/GazzolaLab/PyElastica/pull/405/commits/c5319a6dbc35847af8169dd506ec5ff33ea90f56)

PR #437, opened 2024-10-10
Issue #354 analytical damper fix

### LINKED ISSUE #354: Inconsistent dimensions for damping constants in AnalyticalLinearDamper

sy-cui:
Fix #354.

Instead of the dimensionally inconsistent implementation, three variations of analytical linear dampers are now supported:

1. Uniform damper. The input damping constant `uniform_damping_constant` $\gamma$ is of dimension `1 / T`. Translation and rotation uses the same exponential damping coefficient $\exp(-\gamma t)$. Example:
```python
    simulator.dampen(rod).using(
        AnalyticalLinearDamper,
        uniform_damping_constant=0.1,
        time_step=1e-4, 
    )
```
2. Physical damper. The translational and rotational damping constant are in general different, which follows the following formulae
```math
m\frac{\partial v}{\partial t} = -\gamma_t v, ~~~~ \frac{J}{e}\frac{\partial \omega}{\partial t} = -\gamma_r \omega
```
such that 
```math
[\gamma_t] = FL^{-1} T, ~~~~ [\gamma_r] = FLT
```
The users are responsible for choosing the appropriate, dimensionally consistent damping constants for both modes. Example:
```python
    simulator.dampen(rod).using(
        AnalyticalLinearDamper,
        translational_damping_constant=0.1,
        rotational_damping_constant=0.05,
        time_step=1e-4, 
    )
```

3. Old implementation (to be deprecated). The syntax is designed to be identical as before, so that client codes do not break. Example:
```python
    simulator.dampen(rod).using(
        AnalyticalLinearDamper,
        damping_constant=0.1,
        time_step=1e-4, 
    )
```
Test cases are updated to accommodate the changes. Dissipation module and test cases are moved to dedicated directories.


skim0119 (commented):
Can we keep the changes in `dissipation.py`?

elastica/dissipation/analytical_linear_damper.py
```diff
@@ -0,0 +1,214 @@
...
+
+        return dampen_rates_protocol
+
+    def _physical_damping_protocol(
+        self,
+        translational_damping_constant: np.float64,
+        rotational_damping_constant: np.float64,
+        time_step: np.float64,
+    ) -> DampenType:
+        nodal_mass = self._system.mass.view()
```
armantekinalp, line 195: Does view needed here?

armantekinalp (approved):
LGTM, I have one small comment.

@skim0119 I did not get your point? Can you clarify.

skim0119:
I think changes are good. I'm just asking why we are making separate directory for `dissipation`. I think it is okay to keep them in single file `dissipation.py`, similar to `boundary_condition.py`, `external_forces.py`, etc.

I guess `collision` is exceptional because it requires so many periphery implementation that becomes cluttered, but dissipation implementations are not so complicated at this point.

Unlike CPP or Java, I don't think it is necessary to organize files exclusively class-based hierarchy for python. I'm open for discussion.

elastica/dissipation/analytical_linear_damper.py
sy-cui: No really. Removed.

sy-cui:
@skim0119 Reverted to a single file.

skim0119 (approved):
LGTM

Issue #466, opened 2025-01-31
Numerical noise on `_inv_rotate`


skim0119:
https://github.com/GazzolaLab/PyElastica/blob/d55e86dc0bd1b46b3b8ca88cd460b71392f7c655/elastica/_rotations.py#L152-L154

We add `1e-14` to avoid divide-by-zero error for the equation `x/sin(x)`, which may yield large discrepancy for small `x` (orange):
<img width="596" alt="Image" src="https://github.com/user-attachments/assets/ecc5ee61-70e5-4913-892a-7b372d5fb1d7" />
I've shown with `eps=1e-8` to exaggerate the error.

Numerically, a safer implementation would be `(x+eps)/sin(x+eps)` (blue).

I highly doubt this would change any simulation result, but the numeric noise at zero rotational-angle would be eliminated.

PR #467, opened 2025-02-03
Numerical noise on `_inv_rotate`

Ali-7800:
For issue #466.


skim0119 (commented):
I'll check the example cases if it causes any discrepancy

Issue #507, opened 2025-10-09
Direction of extra segment in knot theory


zzzzzbk:
Hello,

I have a concern about the `def _compute_additional_segment` function inside the [PyElastica](https://github.com/GazzolaLab/PyElastica/tree/aef9e24f4f59b7c54f1028b48332a72dce42604a)/[elastica](https://github.com/GazzolaLab/PyElastica/tree/aef9e24f4f59b7c54f1028b48332a72dce42604a/elastica)/[rod](https://github.com/GazzolaLab/PyElastica/tree/aef9e24f4f59b7c54f1028b48332a72dce42604a/elastica/rod)[/knot_theory.py](https://github.com/GazzolaLab/PyElastica/blob/aef9e24f4f59b7c54f1028b48332a72dce42604a/elastica/rod/knot_theory.py)

    if type_of_additional_segment == "next_tangent":
        for i in range(timesize):
            # Direction of the additional point at the beginning of the rod
            direction_of_rod_begin = center_line[i, :, 0] - center_line[i, :, 1]
            direction_of_rod_begin /= np.linalg.norm(direction_of_rod_begin)

            # Direction of the additional point at the end of the rod
            direction_of_rod_end = center_line[i, :, -1] - center_line[i, :, -2]
            direction_of_rod_end /= np.linalg.norm(direction_of_rod_end)

Here for certain type of additional segment, it loops all timestep to calculate the directions of start/end. but the 'direction_of_rod_end' is not properly saved and it keeps override as i iterates. 

Later when the new points is computed and added, it starts another time loop

    # Compute new centerline and beginning/end direction
    for i in range(timesize):
        first_point = center_line[i, :, 0] + segment_length * direction_of_rod_begin
        last_point = center_line[i, :, -1] + segment_length * direction_of_rod_end

        new_center_line[i, :, 1:-1] = center_line[i, :, :]
        new_center_line[i, :, 0] = first_point
        new_center_line[i, :, -1] = last_point

        beginning_direction[i, :] = direction_of_rod_begin
        end_direction[i, :] = direction_of_rod_end

where the direction_of_rod_begin/end are based on the overridden final state. Also the beginning/end_direction are made from the same vector for all t.

I noticed that because I have a helical curve initially evolves from a straight line. But the link number is non-zero at t=0, altered by the final helical direction. I believe the calculation of link number should be independent to time. 

I can provide some data and simple script to test this if needed. 

Thanks.

PR #509, opened 2025-10-14
fix: compute_additional_segment directions 

zzzzzbk:
Related to #507 

The direction of end/start is not properly saved and causes results affected by the final timestep. 

test_knot_theory: Add a second timestep to the additional segment straight case to test the independence from time.


skim0119:
@zzzzzbk Could you remove the changes from `uv.lock` and `pyproject.toml`?

skim0119 (approved):
LGTM. I'm rebasing the branch; cleaning up some commits.

PR #511, opened 2025-10-18
[Feat] Callback for multiple system types in simulation

### LINKED ISSUE #499: Make Callback-Protocol independant of system (rod/rigid body)
**Is your feature request related to a problem? Please describe.**
1) The current implementaion of the Callback-Protocol follows the `ModuleProtocol` which applies a callback to a specific system (rod/rigid body). When we want to save _all systems_ in a `BaseSystemCollection` into one file, this needs hacking around the protocol. E.g. adding the collection as a parameter to the callback and not using the system provided when the make_callback method is invoked and iterate through all systems in the collection...

2) Additionally the current callbacks are only called at the end of the step, but one does not know if it is the last step in the simulation. This leads to the case when saving every 100th step, it can happen that the solver finishes at some ...99th step. These last 99 steps will never get saved. 

**Describe the solution you'd like**
1) Having an (additional) System-Level-Callback-Protocol on the system level which calls a mehtod e.g. `make_callback(collection: BaseSystemCollection, time: float, step: int)`

2) The protocol shoud include information about the progress of the simulation. Like `make_callback(collection: BaseSystemCollection, time: float, step: int, is_last_step: bool)`. Or call a function like `finished()`, `final_callback()` with the same parameters like `make_callback` (`collection: BaseSystemCollection, time: float, step: int`).

**Describe alternatives you've considered**
Use a workaround like described above E.g. adding the collection as a parameter to the callback...

**Additional context**
We want to save _all systems_ in a `BaseSystemCollection` into _one_ file.

skim0119:
## Feature
- Callback can take `collections` (list, tuple, dict) of system types.
- resolve #499

### DIFF
--- a/docs/guide/workflow.md
+++ b/docs/guide/workflow.md
@@ -221,7 +221,7 @@ simulator.collect_diagnostics(rod2).using(
 You can define different callback functions for different rods and also have different data outputted at different time step intervals depending on your needs. See [this page](../api/callback.rst) for more in-depth documentation.
 
 :::{note}
-During setting up callbacks, `.collect_diagnostics` can python-collection (list, dict, tuple) of rods:
+During setting up callbacks, `.collect_diagnostics` can take python-collection (list, dict, tuple) of rods/systems:
 ```python
 ...
     def make_callback(self, system, time, current_step):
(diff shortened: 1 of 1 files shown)


Ali-7800 (approved):
Tested with list, tuple, dict and ellipsis. LGTM!

PR #513, opened 2025-10-28
[Bug] Fix sphere-rod contact

### LINKED ISSUE #510: _find_min_dist add collision volume for rod sphere contact
**Describe the bug**
I found the rod-sphere contact detection is acting abnormally for certain situation. 

Image
Image

For example in this screenshot, the straight rod is initially have no visual contact with the fixed sphere. But running the simulation will generate a contact force to deform the rod. 

I checked the `class RodSphereContact(NoContact)` the first aabb detects there are overlapping between the rod box and sphere box, but I think it is fine for this very close distance. 

The problem is inside the `_calculate_contact_forces_rod_sphere`.  Noticed when calculate the shortest distanced between rod and sphere, the function `_find_min_dist` for rod-rod contact detection is reused. 
```python
        #find the shortest line segment between the two centerline
        distance_vector, x_sphere_contact_point, _ = _find_min_dist(
            x_selected, edge_collection_rod[..., i], x_sphere_tip, edge_sphere
        )
```

Here it assumes the sphere is a node, starting from the outside of the sphere `x_sphere_tip= sphere_center- sphere_radius* d3` with segment length `edge_sphere = sphere_radius*d3`. Geometrically, that increase the collision volume from a sphere to a cylinder. 

I would suggest replace the the `_find_min_dist` by directly finding the distance from the sphere center to the rod segements. That fixed the contact in my simulation. But I wonder if there is some other reasons for this treatment. 
**To Reproduce**

sphere fixed at center = [ 1.5e-1 -3.91968889e-02 -1.73472348e-18]
rod fixed at the base = [ 0.         -0.0114805  -0.02771639]
rod direction d3=[ 1 0 0]
sphere radius = 0.02
rod radius = 0.01

skim0119:
## Description
- Rod-sphere contact implementation had two bugs: finding distance using cylinder implementation #510, and AABB implementation.

## Minor change
- resolve #510 
- add unittest
- new example case for sphere + rod contact case

## Demo
There should be no contact force, when rod and sphere are exactly touching (no gap).

Before change:

https://github.com/user-attachments/assets/56537650-cad7-4e04-aacf-ad39faff7c9c

After change:

https://github.com/user-attachments/assets/efb2e7f6-b40c-4618-9260-a6aaad471111


examples/RigidBodyCases/RodRigidBodyContact/rod_sphere_contact.py
```diff
@@ -0,0 +1,198 @@
...
+    )
+
+    # add damping
+    damping_constant = 1e-1
+    simulator.dampen(rod).using(
+        ea.AnalyticalLinearDamper,
+        damping_constant=damping_constant,
+        time_step=time_step,
+    )
+    # simulator.dampen(rigid_body).using(
```
Ali-7800: remove this? We don't have damping for rigid bodies

Ali-7800 (approved):
Just minor comment

